PR #5, opened 2019-12-23
fix(vulnerabilities): dependency updates

Francois-Esquire:
Fix for vulnerability (`lodash` version)


prod-sample/sample-modules/frank-lloyd-root/0.0.1/package-lock.json
```diff
@@ -3846,9 +3846,9 @@
       }
     },
     "lodash": {
-      "version": "4.17.11",
-      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.11.tgz",
-      "integrity": "sha1-s56mIp72B+zYniyN8SU2iRysm40="
+      "version": "4.17.15",
+      "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
```
mtomcal: Before we let this through, any chance we can remove this dependency altogether?

Francois-Esquire: To remove `lodash`, we would need to look at the dependencies used in `one-app` and address their usage of the library. `@babel/core`, `@americanexpress/one-app-bundler` and `@americanexpress/eslint-plugin-one-app` are some cases which use `lodash`.

Currently, we are only using `lodash.get` in `prod-sample/sample-modules/preview-frank` and `lodash.set` in `__tests__/client/badPartMonkeypatches.spec.js` as direct dependencies.

mtomcal: I was thinking particularly of `frank-lloyd-root`. The fact that `@americanexpress/one-app-bundler` requires it means we need to keep it. 👍

PR #7, opened 2020-01-01
feat(danger): validate lockfiles

Francois-Esquire:
~~Extends `danger.js` to validate lock files, hosts and protocol of package resolutions are checked.~~
Adding `lockfile-lint` to run on `test:lint` script.


scripts/dangers/keep-package-and-lock-in-sync.js
```diff
@@ -14,6 +14,117 @@
...
+    onSuccess = (...args) => message(...args),
+  }]) => {
+    let result;
+    try {
+      result = validate(schemes);
+    } catch (error) {
+      onWarn(error);
+    }
+
+    if (result.type === 'success') {
```
JAdshead: This will throw a `Cannot read property 'type' of undefined` if `result = validate(schemes);` throws. 

```suggestion
    if (result && result.type === 'success') {
```

scripts/dangers/keep-package-and-lock-in-sync.js
```diff
@@ -14,6 +14,117 @@
...
+export function runLockFilesValidation({ linting = true } = {}) {
+  const cwd = process.cwd();
+  const lockfiles = danger.git.fileMatch('**/package-lock.json');
+
+  if (linting) {
+    const {
+      modified,
+      created,
+      // TODO: pair each package-lock.with package.json, make sure they are not deleted
+      // deleted,
```
JAdshead: duplication of `deleted`

scripts/dangers/keep-package-and-lock-in-sync.js
Francois-Esquire: Good eye!

scripts/dangers/keep-package-and-lock-in-sync.js
Francois-Esquire: So `// deleted,` here is actually a property with the others and I just commented it out as a placeholder with the `TODO`. A reminder. I can remove if that's a blocker.

scripts/dangers/keep-package-and-lock-in-sync.js
```diff
@@ -14,6 +14,117 @@
...
+    let result;
+    try {
+      result = validate(schemes);
+    } catch (error) {
+      onWarn(error);
+    }
+
+    if (result && result.type === 'success') {
+      onSuccess();
+    } else if (result.type === 'error') {
```
JAdshead: also need to check for result here
```suggestion
    } else if (result  && result.type === 'error') {
```

anescobar1991:
Should we just add `lockfile-lint` to the `npm test` script and not rely on danger to run it?

nellyk:
> Should we just add `lockfile-lint` to the `npm test` script and not rely on danger to run it?

I think i would prefer it staying with danger as we are validating whether the package-lock submitted via the pr is valid. It would be still be captured.

Francois-Esquire:
@nellyk @anescobar1991 updated to use command-line tool per team discussion.

package.json
```diff
@@ -22,7 +22,8 @@
     "start:watch": "nodemon --signal SIGTERM --watch src --ext js,jsx --exec babel-node src/server/index.js",
     "prestart:prod-sample": "npm run build:prod-sample",
     "start:prod-sample": "docker-compose -f ./prod-sample/docker-compose.yml up --abort-on-container-exit",
-    "test:lint": "eslint --ext js,jsx,md,snap .",
+    "test:lockfile": "lockfile-lint -p package-lock.json -t npm -a npm -o https: -c -i",
+    "test:lint": "eslint --ext js,jsx,md,snap . && npm run test:lockfile",
```
10xLaCroixDrinker: can you add this to `test` instead of `test:lint`?

Francois-Esquire: Will do that.

Francois-Esquire: updated in 04ca91f

PR #9, opened 2020-01-06
chore(docker-compose): allow for proxy env vars usage in image builds

anescobar1991:
This allows for the integration tests to be ran from within corporate firewalls without any modification being required to the source code or scripts.

Uses https://docs.docker.com/compose/compose-file/#args

Test plan: validated that locally I can run tests with nothing other than setting proxy env vars. CI should validate that everything works when those env vars are excluded as well.


.gitignore
```diff
@@ -27,6 +27,7 @@ prod-sample/nginx/origin-statics
 *.crt
 *.key
 sample-module-bundles
+docker-compose.test.yml
```
anescobar1991: This should have been here all along.

PR #10, opened 2020-01-13
update comments to match valid examples

nellyk:


src/server/config/env/runTime.js
```diff
@@ -118,7 +118,7 @@ const runTime = [
   // locale folder env level
   {
     name: 'ONE_CLIENT_LOCALE_FILENAME',
-    // E1, E2, E3 (E3 use the moduleName)
+    // dev, qa, staging and production (in production use the moduleName)
```
10xLaCroixDrinker: per the line below, `"qa"` and `"integration"` or `undefined` are the valid values. I think this comment may just be removed altogether, unless we make some functionality changes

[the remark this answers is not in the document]
anescobar1991: Ignore that comment, I see @10xLaCroixDrinker already commented about this.

nellyk: thank you, i've updated this

PR #11, opened 2020-01-15
docs(README): update link to frank-lloyd-root partial component

PatNeedham:


anescobar1991:
The travis failure is for the commit message format:
```
⧗   input: docs(README): update link to frank-lloyd-root partial component
```

It doesn't like the scope (README) to be all caps. Don't worry about it we will fix the commit message when we merge.

PR #22, opened 2020-01-27
Feature/add app release

anescobar1991:
Deploy one app and its sample modules to heroku + surge.sh so we can run integration tests against said app.

These integration tests against a deployed instance of One App are only meant to run on commits on master, my thinking is that since these take longer we do not need to run them on PRs. The PRs run integration tests against the docker instance of One App anyways.

This also makes it so that the https://one-app-sample.herokuapp.com/success app is always a representation of `master`.


Dockerfile
```diff
@@ -33,6 +32,7 @@ RUN NODE_ENV=production npm run build && \
 FROM node:12-alpine as development
 ENV NODE_ENV=development \
     HTTP_PORT=3000 \
+    HTTP_ONE_APP_DEV_CDN_PORT=3001 \
```
anescobar1991: Added this in for consistency as the other dev server ports were added.

Dockerfile
```diff
@@ -48,10 +48,7 @@ COPY --from=builder --chown=node:node /opt/one-app/development ./
 # production image
 # last so that it's the default image artifact
 FROM node:12-alpine as production
-ENV NODE_ENV=production \
-    HTTPS_PORT=8443 \
```
anescobar1991: Removing these values as we don't necessarily want to have an opinion on the port numbers our users use and also because by setting them up here prevents the `PORT` env var to override the value from `HTTP_PORT` as required in Heroku and a few other clouds.

__tests__/integration/one-app.spec.js
```diff
@@ -544,6 +544,26 @@ describe('Tests that require Docker setup', () => {
       });
     });
 
+    test('app calls loadModuleData to run async requests using Tenant provided fetchClient', async () => {
```
anescobar1991: Moved this up to the `Docker required` section of the integration tests as this test will never pass when running against a "real" environment because the API servers are not deployed.

src/server/config/env/runTime.js
```diff
@@ -78,11 +88,49 @@ const runTime = [
     },
     defaultValue: () => 3005,
   },
-  // holocron config, the modules and versions to use
+  {
+    name: 'HTTP_ONE_APP_DEV_CDN_PORT',
```
anescobar1991: Added some of these env vars that were not previously added to this file for consistency. As currently implemented there is default env var value logic sprinkled throughout the code and so with this we can centralize that logic.

src/server/listen.js
```diff
@@ -23,7 +23,8 @@ import http from 'http';
 import https from 'https';
 
 function listenHttp(app, cb) {
-  const port = process.env.HTTP_PORT || '3000';
+  // when `undefined` is passed to process.env it sets it to the string undefined...
+  const port = (process.env.HTTP_PORT !== 'undefined' && process.env.HTTP_PORT) || process.env.PORT;
```
anescobar1991: Adding this logic because `PORT` is the env var heroku expects apps to use to know where to listen.

src/server/middleware/sendHtml.js
```diff
@@ -40,8 +40,8 @@ function getChunkAssets(assetsByChunkName) {
     .map(([, assets]) => (typeof assets === 'string' ? assets : assets[0]));
 }
 
-const modernBrowserChunkAssets = getChunkAssets(readJsonFile('../../../.webpack-stats.browser.json').assetsByChunkName);
-const legacyBrowserChunkAssets = getChunkAssets(readJsonFile('../../../.webpack-stats.legacyBrowser.json').assetsByChunkName)
+const modernBrowserChunkAssets = getChunkAssets(readJsonFile('../../../.build-meta.json').modernBrowserChunkAssets);
```
anescobar1991: Made changes in `one-app-bundler` to add all the necessary data previously taken from `webpack-stats-*` files to the `build-meta.json` file. Doing so means we no longer have to add `webpack-stats-*` files to the docker image and prevents memory issues from loading these very large files into memory.

As a result of this change our docker image size went down from ~397mb to ~190mb`.

__tests__/server/listen.spec.js
```diff
@@ -32,6 +32,7 @@ const origEnvVarVals = {};
   'HTTPS_PUBLIC_CERT_CHAIN_PATH',
   'HTTPS_TRUSTED_CA_PATH',
   'HTTPS_PRIVATE_KEY_PASS_FILE_PATH',
+
```
10xLaCroixDrinker:
```suggestion
```

anescobar1991:
**What is the communication plan for this change?**
This is an internal implementation detail and as such does not need to be documented.
**Does any documentation need to be updated or added to account for this change? If so was it done already?**
Already done in this PR.
**What is the motivation for this change?**
To run integration tests against a deployed instance of One App
**Should these changes also be applied to a maintenance branch or any other long lived branch?**
n/a
**How was this change tested?**
https://travis-ci.org/americanexpress/one-app/jobs/642591652?utm_medium=notification&utm_source=email validates running the entire integration test script
**Does this change require cross browser checks? Why or why not?**
n/a
**Does this change require a performance test prior to merging? Why or why not?**
no
**Could this be considered a breaking change? Why or why not?**
no
**Does the change impact caching?**
no
**Does the change impact HTTP headers?**
no
**Does the change have any new infrastructure requirements?**
There were a few environment variables removed as defaults from the dockerfile meaning that those env vars
are now required prior to running One App.
**Does the change affect other versions of the app?**
no
**Does the change require additional environment variables?**
no
**What is the impact to tenants?**
none
**What is the impact to individual users?**
none
**What is the change to the size of assets?**
none, no client side code was touched
**Should integration tests be added to protect against future regressions on this change?**
n/a

__tests__/integration/one-app.spec.js
```diff
@@ -678,7 +698,7 @@ describe('Tests that can run against either local Docker setup or remote One App
             params: {
               0: '/vitruvius',
             },
-            protocol: 'https',
+            protocol: expect.any(String),
```
jsmartfo: reasoning for changing this one?

jsmartfo: just if someone wants to run `http`?

__tests__/server/config/env/runTime.spec.js
```diff
@@ -143,9 +145,92 @@ describe('runTime', () => {
...
+      expect(httpPort.defaultValue()).toBeDefined();
       expect(httpPort.defaultValue()).toBe(3000);
     });
+
+    it('has no default value for production', () => {
+      process.env.NODE_ENV = 'production';
+      expect(httpPort.defaultValue()).not.toBeDefined();
+    });
+
+    it('throws if neither of HTTP_PORT or HTTPS_PORT are not defined', () => {
```
jsmartfo: > it('throws if neither of HTTP_PORT or HTTPS_PORT are not defined', () => {

> it('throws if neither of HTTP_PORT or HTTPS_PORT are defined', () => {

__tests__/integration/one-app.spec.js
PixnBits:
```suggestion
            protocol: expect.toMatch(/https?/);
```

__tests__/integration/one-app.spec.js
```diff
@@ -544,6 +544,26 @@ describe('Tests that require Docker setup', () => {
...
+            author: 'typicode',
+            id: 1,
+            title: 'json-server',
+          },
+        ],
+        secretMessage: 'you are being watched',
+      });
+    });
+
     describe('tentant root configureRequestLog', () => {
```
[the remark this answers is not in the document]
10xLaCroixDrinker: should be removed, rather than fixed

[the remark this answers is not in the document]
anescobar1991: Removed this one and noticed a lot more instances of `tenant` that should be cleaned up. A ticket was opened up internally to track this.

src/server/utils/getI18nFileFromState.js
```diff
@@ -17,7 +17,7 @@
 import readJsonFile from './readJsonFile';
 
 // `require` has a cache, avoid retaining the full file in memory
-const { assetsByChunkName } = readJsonFile('../../../.webpack-stats.browser.json');
+const assetsByChunkName = readJsonFile('../../../.build-meta.json').modernBrowserChunkAssets;
```
PixnBits: could use destructuring?
```suggestion
const { modernBrowserChunkAssets: assetsByChunkName } = readJsonFile('../../../.build-meta.json');
```

src/server/middleware/sendHtml.js
```diff
@@ -40,8 +40,8 @@ function getChunkAssets(assetsByChunkName) {
...
+const legacyBrowserChunkAssets = getChunkAssets(readJsonFile('../../../.build-meta.json').legacyBrowserChunkAssets)
```
PixnBits:
```suggestion
const { modernBrowserChunkAssets } = getChunkAssets(readJsonFile('../../../.build-meta.json'));
const { legacyBrowserChunkAssets } = getChunkAssets(readJsonFile('../../../.build-meta.json'));
```
destructuring might be a style change though?

PixnBits: If you don't go for [destructuring](https://github.com/americanexpress/one-app/pull/22/files#r373077172), be sure to include the semicolon. I'm a bit confused on why linting didn't pick this up, though?
```suggestion
const legacyBrowserChunkAssets = getChunkAssets(readJsonFile('../../../.build-meta.json').legacyBrowserChunkAssets);
```

PixnBits: oh, would be better to combine:
```suggestion
const { modernBrowserChunkAssets, legacyBrowserChunkAssets } = getChunkAssets(readJsonFile('../../../.build-meta.json'));
```

src/server/listen.js
```diff
@@ -23,7 +23,8 @@ import http from 'http';
 import https from 'https';
 
 function listenHttp(app, cb) {
-  const port = process.env.HTTP_PORT || '3000';
+  // when `undefined` is passed to process.env it sets it to the string undefined...
```
PixnBits: prevent assigning `undefined` to an env var, delete it instead

src/server/config/env/runTime.js
```diff
@@ -52,16 +52,26 @@ const runTime = [
...
+        const parsed = parseInt(input, 10);
+        // make sure the parsed value is the same value as input
+        // input may be a string or a number, we don't want === in this case, just ==
+        if (Number.isNaN(parsed) || parsed != input) { // eslint-disable-line eqeqeq
+          throw new Error(`env var HTTP_PORT needs to be a valid integer, given "${input}"`);
+        } else {
+          return parsed;
+        }
       }
+      return undefined;
```
PixnBits: why is this being returned? perhaps normalize shouldn't be called when `input` is falsey (an empty string)?

src/server/config/env/runTime.js
```diff
@@ -52,16 +52,26 @@ const runTime = [
...
     },
-    defaultValue: () => 3000,
+    validate: (value) => {
+      if (!process.env.HTTPS_PORT && !value && !process.env.PORT) {
+        throw new Error('Either of the `HTTP_PORT` or `HTTPS_PORT` or `PORT` environment variables must be defined');
+      }
+    },
+    defaultValue: () => (process.env.NODE_ENV === 'development'
+      ? 3000
+      : undefined),
```
PixnBits: `undefined` gets coerced to a string in `process.env` as you found, let's rethink this approach? what are the intentions behind this?

Dockerfile
PixnBits: these pair with the `EXPOSE` statements below https://docs.docker.com/engine/reference/builder/#expose
hm 🤔

__tests__/server/config/env/__snapshots__/runTime.spec.js.snap
```diff
@@ -1,5 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`runTime HTTP_ONE_APP_DEV_CDN_PORT normalizes numeric input 1`] = `"env var HTTP_ONE_APP_DEV_CDN_PORT needs to be a valid integer, given \\"0002345a\\""`;
+
+exports[`runTime HTTP_ONE_APP_DEV_PROXY_SERVER_PORT normalizes numeric input: env var HTTP_ONE_APP_DEV_PROXY_SERVER_PORT needs to be a valid integer, given "r00t" 1`] = `"env var HTTP_ONE_APP_DEV_PROXY_SERVER_PORT needs to be a valid integer, given \\"r00t\\""`;
+
+exports[`runTime HTTP_PORT throws if neither of HTTP_PORT or HTTPS_PORT are defined 1`] = `"Either of the \`HTTP_PORT\` or \`HTTPS_PORT\` or \`PORT\` environment variables must be defined"`;
```
PixnBits: "Either" means two choices I think, but the message has three
```suggestion
exports[`runTime HTTP_PORT throws if neither of HTTP_PORT or HTTPS_PORT are defined 1`] = `"One of the \`HTTP_PORT\` or \`HTTPS_PORT\` or \`PORT\` environment variables must be defined"`;
```

__tests__/integration/README.md
```diff
@@ -54,13 +54,15 @@ To make it easier for you to deploy all the sample modules to a remote environme
 to build and bundle the modules and module map for you:
 
 ```bash
-npm run build:sample-modules -- --archiveBuiltArtifacts
+npm run build:sample-modules -- --archive-built-artifacts --bundle-statics-hostname=https://cdn.example.com
```
PixnBits: strictly speaking `https://cdn.example.com` is more than a [hostname](https://en.wikipedia.org/wiki/Hostname); from [a diagram in Node.JS's URL documentation](https://nodejs.org/dist/latest-v10.x/docs/api/url.html#url_url_strings_and_url_objects) this appears to be an `origin`

Dockerfile
PixnBits: these pair with the [EXPOSE](https://docs.docker.com/engine/reference/builder/#expose) statements below
>[`EXPOSE`] functions as a type of documentation between the person who builds the image and the person who runs the container, about which ports are intended to be published.

The image should have some defaults IMO but I was also under the impression that each of the env vars could be overridden, including removal. Does Heroku support removing env vars? Does that look like `HTTP_PORT=''`?
hm 🤔
this might be a case where Docker and Heroku disagree about best practice, the downside of following Docker's is the added work/friction for devs to get running on Heroku, the downside of following Heroku's is anything docker-native? but IDK if the hub even uses this information in the UI? ex: https://hub.docker.com/_/nginx

src/server/config/env/runTime.js
```diff
@@ -52,16 +52,26 @@ const runTime = [
...
       }
+      return undefined;
     },
-    defaultValue: () => 3000,
+    validate: (value) => {
+      if (!process.env.HTTPS_PORT && !value && !process.env.PORT) {
+        throw new Error('Either of the `HTTP_PORT` or `HTTPS_PORT` or `PORT` environment variables must be defined');
+      }
+    },
+    defaultValue: () => (process.env.NODE_ENV === 'development'
```
PixnBits: use the value of `PORT` here in `defaultValue`
I believe this also removes the need for returning `undefined` from the `normalize` function

src/server/config/env/runTime.js
```diff
@@ -78,11 +88,49 @@ const runTime = [
...
+        const parsed = parseInt(input, 10);
+        // make sure the parsed value is the same value as input
+        // input may be a string or a number, we don't want === in this case, just ==
+        if (Number.isNaN(parsed) || parsed != input) { // eslint-disable-line eqeqeq
+          throw new Error(`env var HTTP_ONE_APP_DEV_CDN_PORT needs to be a valid integer, given "${input}"`);
+        } else {
+          return parsed;
+        }
+      }
+      return undefined;
```
PixnBits: `undefined` is coerced to a string in `process.env` as you found
does env-config-utils preprocessEnvVar.js support optional configuration?
(might need to add that? ex:
```javascript
// https://github.com/americanexpress/env-config-utils/blob/master/src/preprocessEnvVar.js#L74
// from 
process.env[name] = value;
// to
if (value === undefined) {
  delete process.env[name];
} else {
  process.env[name] = value;
}
```
but then, perhaps `normalize` should just return the original input instead? maybe we should open an issue in that repo to discuss the feature requirements and implications in detail?)

src/server/listen.js
PixnBits: Took me too long 😓but I'm now realizing that should be done in `defaultValue` of `src/server/config/env/runTime.js`.
```suggestion
  const port = process.env.HTTP_PORT;
```

prod-sample/sample-modules/frank-lloyd-root/0.0.0/src/csp.js
```diff
@@ -40,9 +48,14 @@ export default contentSecurityPolicyBuilder({
     connectSrc: [
       "'self'",
       '*.api.frank',
-      `${ip.address()}:${process.env.HTTP_ONE_APP_DEV_CDN_PORT || 3001}`,
-      `localhost:${process.env.HTTP_ONE_APP_DEV_CDN_PORT || 3001}`,
+      // used for local development
+      `${ip.address()}:3001`,
```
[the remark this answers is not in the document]
anescobar1991: If this variable is undefined (which it is in any production set up) then the CSP is invalid.

I think with the normalize updates you mentioned this will now work as intended. The issue was that the string `undefined` was used as the value to `HTTP_ONE_APP_DEV_CDN_PORT` and hence the logic for default port 3001 was never reached.

scripts/build-sample-modules.js
```diff
@@ -204,33 +206,11 @@ const doWork = async () => {
...
-      { cwd: nginxOriginStaticsRootDir }
-    );
 
-    // eslint has conflicting rules
-    // eslint-disable-next-line max-len
-    await fs.move(path.join(nginxOriginStaticsRootDir, moduleMapArchiveName), path.join(pathToBundles, moduleMapArchiveName), { overwrite: true });
+    Promise.all([
+      await fs.move(nginxOriginStaticsModulesDir, path.join(pathToBundles, 'modules'), { overwrite: true }),
+      await fs.move(pathToNginxOriginModuleMap, path.join(pathToBundles, 'module-map.json'), { overwrite: true }),
+    ]);
```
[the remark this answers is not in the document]
anescobar1991: 🤦‍♂ yeah you're right.

scripts/build-sample-modules.js
```diff
@@ -35,7 +36,8 @@ const nginxOriginStaticsModulesDir = path.resolve(nginxOriginStaticsRootDir, 'mo
 const pathToNginxOriginModuleMap = path.resolve(nginxOriginStaticsRootDir, 'module-map.json');
 const pathToAssets = path.resolve(sampleProdDir, 'assets');
 const userIntendsToSkipSampleModulesBuild = process.env.ONE_DANGEROUSLY_SKIP_SAMPLE_MODULES_BUILD;
-const archiveBuiltArtifacts = process.argv.includes('--archiveBuiltArtifacts');
+const archiveBuiltArtifacts = argv.archiveBuiltArtifacts || '';
+const bundleStaticsHostname = argv.bundleStaticsHostname || 'https://sample-cdn.frank';
```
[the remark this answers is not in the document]
anescobar1991: there is no `yargs` config for this script as it is a standalone script and I didnt want to add that overhead. The `yargs` config you are thinking about is the one for One App itself.

__tests__/server/config/env/__snapshots__/runTime.spec.js.snap
anescobar1991: Per webster either can mean 2 or sometimes more than 2.

src/server/middleware/sendHtml.js
anescobar1991: I'll look into why linting didnt pick this up. That is weird.

Dockerfile
anescobar1991: No they can't really be removed. This is what heroku's docs have to say about EXPOSE.

https://devcenter.heroku.com/articles/container-registry-and-runtime#dockerfile-commands-and-runtime

EXPOSE is a _suggestion_ per docker docs so I think the right thing to do here is remove EXPOSE as well from the dockerfile.

__tests__/server/listen.spec.js
```diff
@@ -47,6 +47,7 @@ function clearAllEnvVars() {
   Object
     .keys(origEnvVarVals)
     .forEach((name) => delete process.env[name]);
+  process.env.HTTP_PORT = 3000;
```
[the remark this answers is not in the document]
anescobar1991: I should probably rename it to `resetAllEnvVars`?

__tests__/server/listen.spec.js
```diff
@@ -112,16 +113,17 @@ describe('server listen', () => {
       expect(cb).toHaveBeenCalledWith(null, { port: '8998' });
     });
 
-    it('uses a default port if HTTP_PORT env var not given', () => {
```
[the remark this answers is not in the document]
anescobar1991: But it doesn't though when `NODE_ENV` is set to `production`.

src/server/middleware/sendHtml.js
anescobar1991: The reason as to the semicolon is omitted is bc that line in question is part of:

```js
const legacyBrowserChunkAssets = getChunkAssets(readJsonFile('../../../.build-meta.json').legacyBrowserChunkAssets)
  .map((chunkAsset) => `legacy/${chunkAsset}`);
```

anescobar1991: And the reason for no destructuring is that `readJsonFile` is what is returning the object that includes the chunk assets. See the parentheses.

anescobar1991: See https://github.com/americanexpress/one-app/pull/22#discussion_r373794000

scripts/build-sample-modules.js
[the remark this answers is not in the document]
anescobar1991: This actually revealed another bug in the line immediately above:  `fs.emptyDir(pathToBundles);` should be `await fs.emptyDir(pathToBundles);`

__tests__/integration/README.md
anescobar1991: Resolved in latest commit.

scripts/build-sample-modules.js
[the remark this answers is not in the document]
anescobar1991: Resolved in latest commit.

src/server/utils/getI18nFileFromState.js
anescobar1991: Resolved in latest commit.

__tests__/server/config/env/runTime.spec.js
```diff
@@ -170,7 +255,7 @@ describe('runTime', () => {
 
     it('has a default value for development', () => {
       process.env.NODE_ENV = 'development';
-      delete process.env.HTTP_ONE_APP_DEV_CDN_PORT;
+      process.env.HTTP_ONE_APP_DEV_CDN_PORT = 3001;
```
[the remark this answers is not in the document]
anescobar1991: It doesn't bc this test is for `HOLOCRON_MODULE_MAP_URL` and the default value is "`http://${ip}:${process.env.HTTP_ONE_APP_DEV_CDN_PORT}/static/module-map.json`"

anescobar1991:
@PixnBits I made the changes per your review and responded to some of your comments.

As for the runtime changes I couldn't figure out a way to make the changes you are suggesting without making breaking changes to `runtime-config-utils`. Maybe we need to catch up on this on Monday?

src/server/middleware/sendHtml.js
PixnBits: might save memory and CPU if the file is read and parsed only once, IDK when the garbage collection would run after the first parse so two copies might still be in memory at that point

__tests__/integration/README.md
PixnBits:
```suggestion
npm run build:sample-modules -- --archive-built-artifacts --bundle-statics-origin=https://cdn.example.com
```

scripts/build-sample-modules.js
```diff
@@ -35,7 +36,8 @@ const nginxOriginStaticsModulesDir = path.resolve(nginxOriginStaticsRootDir, 'mo
 const pathToNginxOriginModuleMap = path.resolve(nginxOriginStaticsRootDir, 'module-map.json');
 const pathToAssets = path.resolve(sampleProdDir, 'assets');
 const userIntendsToSkipSampleModulesBuild = process.env.ONE_DANGEROUSLY_SKIP_SAMPLE_MODULES_BUILD;
-const archiveBuiltArtifacts = process.argv.includes('--archiveBuiltArtifacts');
+const archiveBuiltArtifacts = argv.archiveBuiltArtifacts || '';
+const bundleStaticsOrigin = argv.bundleStaticsOrigin || 'https://sample-cdn.frank';
```
PixnBits: should `'https://sample-cdn.frank'` be the default set up in [`src/server/config/env/argv.js`](https://github.com/americanexpress/one-app/blob/master/src/server/config/env/argv.js)? (similar for `archiveBuiltArtifacts`)
```suggestion
const bundleStaticsOrigin = argv.bundleStaticsOrigin;
```

anescobar1991: This script is not a part of One App, it is a standalone script used to build the sample modules to run integration tests against. The set up in `src/server/config/env/argv.js` only applies to One App's node process.

src/server/config/env/runTime.js
anescobar1991: This should be done in a separate effort since it involves changes to another library that are out of scope for this PR.

anescobar1991:
@PixnBits I _think_I made the changes you suggested! It feels much cleaner now.

__tests__/server/index.spec.js
```diff
@@ -137,13 +137,6 @@ describe('server index', () => {
       expect(yargs.getOptions().string).toMatchSnapshot();
     });
 
-    it('starts devHolocronCDN on port 3001 when no env variable is set', async () => {
```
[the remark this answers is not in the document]
anescobar1991: This logic is now outside this file and thus is validated in the unit tests for the runtime.js file.

__tests__/server/config/env/__snapshots__/runTime.spec.js.snap
PixnBits: depends on the usage
>Definition of either
>
>adjective
1: being the one and the other of two : EACH
_flowers blooming on either side of the walk_
_plays either instrument well_
2: being the one or the other of two
_take either road_
>
>pronoun
the one or the other
_take either of the two routes_
>
>conjunction
used as a function word before two or more coordinate words, phrases, or clauses joined usually by or to indicate that what immediately follows is the first of two or more alternatives
_can be used either as a guest room or as an office_
>
>adverb
1: LIKEWISE, MOREOVER —used for emphasis after a negative
_not smart or handsome either_
2: for that matter —used for emphasis after an alternative following a question or conditional clause especially where negation is implied
_who answers for the Irish parliament? or army either? — Robert Browning_

https://www.merriam-webster.com/dictionary/either

The conjunction form is the only one that mentions more than 2, and that's after saying 2 which appears to be the more common/preferred usage. I think that's due to [writers ignoring "the rules"](https://www.grammarphobia.com/blog/2017/08/either-neither.html) rather than [common usage](https://ell.stackexchange.com/a/63450):
ngram usage of "either of the three" and "any of the three" with the former dropping significantly since 1880

I think I read something about formal (2) vs informal/conversational (2 & 2+) rules but can't find that now. IMO it's better for technical messages to adhere to formal rules to avoid ambiguity. 

At the risk of bike-shedding I won't require a change here, but I would ask to strongly consider the point on ambiguity.

Thanks, @anescobar1991 !

src/server/config/env/runTime.js
PixnBits: My only concern after the latest changes are having `process.env.HTTP_ONE_APP_DEV_CDN_PORT === 'undefined'`, but perhaps that's fine for now as something attempting to start the dev CDN server in production mode should fail anyway.

__tests__/server/config/env/__snapshots__/runTime.spec.js.snap
PixnBits: I was viewing an older diff, this isn't even an issue after the latest changes 🤦‍♂

PR #28, opened 2020-02-04
test(integration): reporting routes

JAdshead:


10xLaCroixDrinker:
You just need to  
- [ ] address the preflight checklist

JAdshead:
N/A tests
N/A
increase integration test coverage
NA
How was this change tested?
These are tests
NO, tests only
NO, Tests only
NO
NO
NO
NO
No
No
What is the impact to tenants?
none
What is the impact to individual users?
none
none
no, these are integration tests

PR #30, opened 2020-02-08
docs(state-management): clear up

anescobar1991:
Cleared up a few points to make docs friendlier to new users

### DIFF
--- a/docs/api/modules/State-Management.md
+++ b/docs/api/modules/State-Management.md
@@ -69,7 +69,8 @@ The following API definitions describe the Ducks responsible for the state shape
 * [`intl`](#intl-duck)
 
 ### `config` Duck
-The `config` Duck lists a subset of the environment variables set on the Server.
+The `config` Duck lists a subset of the environment variables set on the Server as well as all values set by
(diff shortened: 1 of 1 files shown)


10xLaCroixDrinker:
@anescobar1991 please fill out the preflight checklist

anescobar1991:
*What is the communication plan for this change?*
n/a
*Does any documentation need to be updated or added to account for this change? If so was it done already?*
This is a docs PR
*What is the motivation for this change?*
To clarify the shape of the state config provided one app to new users
*Should these changes also be applied to a maintenance branch or any other long lived branch?*
n/a
*How was this change tested?*
viewed the preview in github to make sure it renders as expected
*Does this change require cross browser checks? Why or why not?*
no
*Does this change require a performance test prior to merging? Why or why not?*
no
*Could this be considered a breaking change? Why or why not?*
no bc this change only affects docs
*Does the change impact caching?*
no
*Does the change impact HTTP headers?*
no
*Does the change have any new infrastructure requirements?*
no
*Does the change affect other versions of the app?*
no
*Does the change require additional environment variables?*
no
*What is the impact to tenants?*
easier for them to figure out how to use One App
*What is the impact to individual users?*
none
*What is the change to the size of assets?*
none
*Should integration tests be added to protect against future regressions on this change?*
overkill

PR #32, opened 2020-02-10
fix(provided-externals): logs warning for child modules

Francois-Esquire:
Replaced error thrown `onModuleLoad` with warning log for child modules that contain the `providedExternals` in `appConfig`.


prod-sample/sample-modules/late-frank/0.0.1/src/index.js
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import ModuleContainer from './components/LateFrank';
+
+if (!global.BROWSER) {
+  ModuleContainer.appConfig = {
```
10xLaCroixDrinker: this should be done by the bundler

Francois-Esquire:
☑️ Preflight Checklist:

None, `providedExternals` pending documentation

No

To prevent the app from crashing based on child module mis-configuration

No

How was this change tested?
Unit and integration testing

No, server side behavior

No, replaced throwing an error with `console.warn`

No, API is unchanged

No

No

No

No

No

What is the impact to tenants?
None

What is the impact to individual users?
Warning on mis-configuration of `appConfig`

None

Added

__tests__/integration/one-app.spec.js
```diff
@@ -568,6 +568,41 @@ describe('Tests that require Docker setup', () => {
...
+          // not ideal but need to wait for app to poll;
+          await waitFor(5000);
+        });
+
+        afterAll(() => {
+          writeModuleMap(originalModuleMap);
+        });
+
+        test(
+          'writes a warning to log if a child module has set `providedExternals` in `appConfig`',
```
10xLaCroixDrinker: you need to test that the externals provided by the root are still correct/available and that the ones provided by the non-root module are not available

src/server/utils/onModuleLoad.js
```diff
@@ -118,7 +118,9 @@ export default function onModuleLoad({
   }
 
   if (providedExternals) {
-    throw new Error(`Module ${moduleName} attempted to provide externals, but it is not the root module.`);
+    console.warn(
+      `Module ${moduleName} attempted to provide externals. Only the root module can provide externals.`
+    );
```
10xLaCroixDrinker: You will need some changes around here: https://github.com/americanexpress/one-app-cli/blob/master/packages/one-app-bundler/webpack/loaders/provided-externals-loader.js#L34

__tests__/integration/one-app.spec.js
Francois-Esquire: Got it, I'll update the tests

prod-sample/sample-modules/late-frank/0.0.1/src/index.js
Francois-Esquire: What do you mean by this specifically? The configuration?

Francois-Esquire: I see what you mean now, will update.

src/server/utils/onModuleLoad.js
Francois-Esquire: I think it's best we move that helper added by the loader into One App -> `getRootModule` and remove it from being injected. Thoughts?

mtomcal: - [ ] Get @10xLaCroixDrinker Feedback on this

__tests__/integration/one-app.spec.js
Francois-Esquire: updated aa0b3a1

prod-sample/sample-modules/late-frank/0.0.1/src/index.js
Francois-Esquire: updated aa0b3a1

PR #33, opened 2020-02-10
feat(routes): added default 404 route

10xLaCroixDrinker:


10xLaCroixDrinker:
- What is the communication plan for this change?
  Inclusion in release notes
- Does any documentation need to be updated or added to account for this change? If so was it done already?
  no
- What is the motivation for this change?
  handle route match misses. previously, going through a bad link would keep you on the same page with an error logged from the router. if the user then refreshed they would see the express 404 page. at that point using the3 back button would not work
- Should these changes also be applied to a maintenance branch or any other long lived branch?
  no
- How was this change tested?
  unit tests, integration test, manual validation with sample modules.
- Does this change require cross browser checks? Why or why not?
  no, just a route addition
- Does this change require a performance test prior to merging? Why or why not?
- Could this be considered a breaking change? Why or why not?
  no, it is only additive. users can still define their own 404 handler
- Does the change impact caching?
  no
- Does the change impact HTTP headers?
  no
- Does the change have any new infrastructure requirements?
  no
- Does the change affect other versions of the app?
  no
- Does the change require additional environment variables?
  no
- What is the impact to tenants?
  gracefully handled 404s
- What is the impact to individual users?
  none
- What is the change to the size of assets?
  none
- Should integration tests be added to protect against future regressions on this change?
  added

PR #35, opened 2020-02-11
test(integration): domain scanning

JAdshead:


JAdshead:
NA
Yes, has been included
Provide way to test user arriving at different versions of one-app
yes, done
How was this change tested?
locally, red - green
no, tests only
no,test only
no, tests only
no
no
no
no
no
What is the impact to developers using the app?
none
none
no

PR #38, opened 2020-02-13
feat(stateConfig): make state config more flexible

### LINKED ISSUE #20: The config from the root module should be more flexible
The env vars are parsed, so they are not always strings

https://github.com/americanexpress/one-app/blob/master/src/server/utils/stateConfig.js#L23-L37

The config from the root module should be more flexible

_Originally posted by @10xLaCroixDrinker in https://github.com/americanexpress/one-app/pull/19/review_comment/create_

JamesSingleton:
This allows the provideStateConfig to accept more than an object
or a string.

Closes #20

### DIFF
--- a/docs/api/modules/App-Configuration.md
+++ b/docs/api/modules/App-Configuration.md
@@ -34,15 +34,16 @@ Node Bundle (e.g.`mymodule.node.js`) rather than the Browser Bundles (e.g.
 security and bundle size considerations.
 
 **Contents**
-* [provideStateConfig](#providestateconfig)
-* [csp](#csp)
-* [corsOrigins](#corsorigins)
-* [configureRequestLog](#configurerequestlog)
-* [extendSafeRequestRestrictedAttributes](#extendsaferequestrestrictedattributes)
-* [createSsrFetch](#createssrfetch)
-* [validateStateConfig](#validatestateconfig)
-* [requiredSafeRequestRestrictedAttributes](#requiredsaferequestrestrictedattributes)
-* [appCompatibility](#appcompatibility)
+- [App Configuration](#app-configuration)
+  - [`provideStateConfig`](#providestateconfig)
+  - [`csp`](#csp)
+  - [`corsOrigins`](#corsorigins)
+  - [`configureRequestLog`](#configurerequestlog)
+  - [`extendSafeRequestRestrictedAttributes`](#extendsaferequestrestrictedattributes)
+  - [`createSsrFetch`](#createssrfetch)
+  - [`validateStateConfig`](#validatestateconfig)
+  - [`requiredSafeRequestRestrictedAttributes`](#requiredsaferequestrestrictedattributes)
+  - [`appCompatibility`](#appcompatibility)
 
 ## `provideStateConfig`
 **Module Type**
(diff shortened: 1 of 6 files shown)


JamesSingleton:
- What is the communication plan for this change?
  - API Docs
- Does any documentation need to be updated or added to account for this change? If so was it done already?
  - Potentially https://github.com/americanexpress/one-app/blob/master/docs/api/modules/App-Configuration.md#providestateconfig.
  - No it has not been updated, not sure if it is needed.
- What is the motivation for this change?
  - #20 
- Should these changes also be applied to a maintenance branch or any other long lived branch?
  - No
- How was this change tested?
  - Unit tests
  - Manually serving a root module to `one-app`
- Does this change require cross browser checks? Why or why not?
  - N/A
- Does this change require a performance test prior to merging? Why or why not?
  - N/A
- Could this be considered a breaking change? Why or why not?
  - No, relaxing the check on `provideStateConfig`
- Does the change impact caching?
  - No
- Does the change impact HTTP headers?
  - No
- Does the change have any new infrastructure requirements?
  - No
- Does the change affect other versions of the app?
  - No
- Does the change require additional environment variables?
  - No
- What is the impact to developers using the app?
  - Allows developers to provide a more flexible `provideStateConfig`
- What is the change to the size of assets?
  - N/A
- Should integration tests be added to protect against future regressions on this change?
  - Not sure, would like some input from @anescobar1991 @10xLaCroixDrinker @Francois-Esquire @nellyk, @mtomcal and @JAdshead on this

docs/api/modules/App-Configuration.md
```diff
@@ -34,15 +34,16 @@ Node Bundle (e.g.`mymodule.node.js`) rather than the Browser Bundles (e.g.
...
-* [provideStateConfig](#providestateconfig)
-* [csp](#csp)
-* [corsOrigins](#corsorigins)
-* [configureRequestLog](#configurerequestlog)
-* [extendSafeRequestRestrictedAttributes](#extendsaferequestrestrictedattributes)
-* [createSsrFetch](#createssrfetch)
-* [validateStateConfig](#validatestateconfig)
-* [requiredSafeRequestRestrictedAttributes](#requiredsaferequestrestrictedattributes)
-* [appCompatibility](#appcompatibility)
+- [App Configuration](#app-configuration)
```
10xLaCroixDrinker: Why do we need this?

src/server/utils/stateConfig.js
```diff
@@ -123,8 +123,7 @@ export const setStateConfig = (providedStateConfig) => {
             throw makeConfigEnvError();
           }
           acc.client[key] = client[configEnv];
-        }
-        if (typeof client === 'string') {
+        } else {
```
10xLaCroixDrinker: keep validating the type

JamesSingleton: So we want to only explicitly accept string, number, or boolean? The reason I did this was it puts the complexity to 13. I talked with @JAdshead and we came up with this solution.

docs/api/modules/App-Configuration.md
JamesSingleton: I don't remember adding this... Let me remove it

JamesSingleton:
> Looks great to me. We should add a test case to Integration test to see if non string is parsed correctly as well.

@mtomcal,

This would require a `0.0.2` of PickyFrank to be created since the current way we log things, it throws at the first error instead of batching everything together.

PR #39, opened 2020-02-24
docs(readme): update readme for generator

nellyk:
Added NODE_ENV as it is required for the one-app-generator to create a module

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ Want to get paid for your contributions to `one-app`?
 The easiest way to do this is via [`npx`](https://blog.npmjs.org/post/162869356040/introducing-npx-an-npm-package-runner) (comes with `npm` versions 5.2.0 and above). Run the following command in the directory you want your module to live:
 
 ```bash
+export NODE_ENV=development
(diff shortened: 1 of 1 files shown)


10xLaCroixDrinker:
Commit type should be `docs`

nellyk:
What is the communication plan for this change?N/A
Does any documentation need to be updated or added to account for this change? If so was it done already?N/A
What is the motivation for this change? For ensure correct installation when creating one-app modules
How was this change tested?N/A
Does this change require cross browser checks? Why or why not?N/A
Does this change require a performance test prior to merging? Why or why not?N/A
Could this be considered a breaking change? Why or why not?Nope, updating the docs
Does the change impact caching?N/A
Does the change impact HTTP headers?N/A
Does the change have any new infrastructure requirements?N/A
Does the change affect other versions of the app?N/A
Does the change require additional environment variables?N/A
What is the impact to developers using the app?N/A
What is the change to the size of assets?N/A
Should integration tests be added to protect against future regressions on this change?N/A

anescobar1991:
Good catch @nellyk!

PR #37, opened 2020-02-13
feat(loadModule): optional cache bust key

JAdshead:
- [x]  include holocron update

PR #40, opened 2020-02-24
docs(server): added module map and env var docs

mtomcal:
- Added documentation on Module Map Schema https://github.com/americanexpress/one-app/blob/docs/module-map-schema/docs/api/server/Module-Map-Schema.md
- Added reformatted and expanded on Environment Variables documentation https://github.com/americanexpress/one-app/blob/docs/module-map-schema/docs/api/server/Environment-Variables.md


docs/api/server/Module-Map-Schema.md
```diff
@@ -1,7 +1,109 @@
...
+  * [`legacyBrowser` Field](#legacyBrowser-field)
+  * [`node` Field](#node-field)
+  * [`url` Field](#url-field)
+  * [`integrity` Field](#integrity-field)
+
+## Shape
+
+```js
+({
+  key: String, // required
```
anescobar1991: With #37 this is optional so should annotate as such here.

docs/api/server/Module-Map-Schema.md
```diff
@@ -1,7 +1,109 @@
...
+    },
+    // ... more module entries allowed
+  },
+});
+```
+
+## Example
+```json
+{
+  "key": "<redacted>",
```
anescobar1991: Why `<redacted>`?  This makes it seem like it is secret in some way when really it could be any string.

docs/api/server/Environment-Variables.md
```diff
@@ -2,6 +2,461 @@
...
+
+## `HOLOCRON_MODULE_MAP_URL`
+
+**Runs In**
+* ✅ Production
+* ✅ Development
+
+URL where the [Module Map] is hosted.
+
+Defaults to `http://localhost:3001/static/module-map.json` if `NODE_ENV` is set to `development` so that [One App Dev CDN] can be leveraged for local development.
```
anescobar1991: Do you think it would be more clear to readers if the default value was given its own section instead of it being hidden in the description where it may go unnoticed?

docs/api/server/Environment-Variables.md
```diff
@@ -2,6 +2,461 @@
...
+* [`HTTPS_PRIVATE_KEY_PASS_FILE_PATH`](#https_private_key_pass_file_path)
+* [`HTTPS_PRIVATE_KEY_PATH`](#https_private_key_path)
+* [`HTTPS_PUBLIC_CERT_CHAIN_PATH`](#https_public_cert_chain_path)
+* [`HTTPS_TRUSTED_CA_PATH`](#https_trusted_ca_path)
+* [`HTTP_METRICS_PORT`](#http_metrics_port)
+* [`HTTP_ONE_APP_DEV_CDN_PORT`](#http_one_app_dev_cdn_port)
+* [`HTTP_ONE_APP_DEV_PROXY_SERVER_PORT`](#http_one_app_dev_proxy_server_port)
+* [`HTTP_PORT`](#http_port)
+* [`IP_ADDRESS`](#ip_address)
+* [`NODE_ENV`](#node_env)
```
anescobar1991: Should we order these by most likely to be used? For example `NODE_ENV` would go at the top since it always has to be set. Or maybe group by development only env vars vs production ones vs ones applicable both in dev and prod?

docs/api/server/Environment-Variables.md
```diff
@@ -2,6 +2,461 @@
...
+```bash
+HTTPS_PUBLIC_CERT_CHAIN_PATH=String
+```
+
+**Example**
+```bash
+HTTPS_PUBLIC_CERT_CHAIN_PATH=./some-cert.pem
+```
+
+## `HTTPS_TRUSTED_CA_PATH`
```
anescobar1991: For each env var can we define if they are required and if so in what situations?

For example this env var I believe is only required if `HTTPS_PORT` is provided.

docs/api/server/Environment-Variables.md
```diff
@@ -2,6 +2,461 @@
 
 # Environment Variables
 
-Documentation Forthcoming
+One App can be configured via Environment Variables:
 
-[☝️ Return To Top](#environment-variables)
\ No newline at end of file
+**Contents**
```
nellyk: should we also talk about how you can start setting this up where do you run them and when? do you have to restart once you this is setup? what happens if you need to change any of this?

mtomcal: @PixnBits Any ideas on grouping ENV vars from an Ops perspective?

docs/api/server/Environment-Variables.md
```diff
@@ -2,6 +2,461 @@
...
+NODE_ENV=production
+```
+
+## `ONE_CLIENT_CDN_URL`
+
+**Runs In**
+* ✅ Production
+* ✅ Development
+
+Fully qualified base path URL where the Module Bundle folders are located (See [Module Map]). Defaults to `/_/static/`.
```
infoxicator: Last time I tried this envVar was required in production and didn't default to `/_/static/` was that changed? I couldn't find the PR

mtomcal: Fixed in b02bcb1

docs/api/server/Environment-Variables.md
mtomcal: Added a topic based ToC in b02bcb1

docs/api/server/Environment-Variables.md
mtomcal: I created a By Topic ToC in b02bcb1

docs/api/server/Environment-Variables.md
mtomcal: Updated in b02bcb1

docs/api/server/Module-Map-Schema.md
mtomcal: Fixed in b02bcb1

docs/api/server/Module-Map-Schema.md
mtomcal: Fixed in b02bcb1

docs/api/server/Environment-Variables.md
mtomcal: Fixed in 3aa1d50

docs/api/server/Environment-Variables.md
```diff
@@ -2,6 +2,542 @@
 
 # Environment Variables
 
-Documentation Forthcoming
+One App can be configured via Environment Variables:
 
-[☝️ Return To Top](#environment-variables)
\ No newline at end of file
+**By Topic**
```
JamesSingleton: Shouldn't this be `## 📖 Table of Contents`

JamesSingleton:
One thing I noticed, is we say `Integer` instead of `Number`. If I remember correctly an Integer represents a floating point number. Also, `Integer` is not one of the second primitive data types.

docs/api/server/Environment-Variables.md
mtomcal: Fixed in cf746a2

mtomcal:
> One thing I noticed, is we say `Integer` instead of `Number`. If I remember correctly an Integer represents a floating point number. Also, `Integer` is not one of the seven primitive data types.

Fixed in cf746a2

PR #41, opened 2020-02-25
feat(html-shell): html and script rendering utils

Francois-Esquire:
Changes to move some rendering logic outside of `sendHtml` and to utils. Rendered html is minified with human readable breaks on state.


JAdshead:
@Francois-Esquire can you break this up into multiple PRs ?

PR #42, opened 2020-02-25
feat(csp): add local development domains to csp

stephaniecoates:
moves logic to add IP and `localhost` to csp when `NODE_ENV=development` from root module to One App.

One thing to note: root module _must_ have csp directives (`script-src`, `connect-src`) defined already in order for ip and localhost to be added. One App only adds the values to `script-src` and `connect-src` if they're pre-existing directives.

Feedback on the explanation in the docs is much appreciated.

### DIFF
--- a/__tests__/server/middleware/csp.spec.js
+++ b/__tests__/server/middleware/csp.spec.js
@@ -99,9 +99,9 @@ describe('csp', () => {
       const headers = res._getHeaders();
       expect(headers).toHaveProperty('content-security-policy');
       const cspString = headers['content-security-policy'];
-      const ipFound = cspString.match(/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}:3001/);
+      const ipFound = cspString.match(/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/);
       expect(ipFound).toBeNull();
-      const localhostFound = cspString.match(/localhost:3001/);
+      const localhostFound = cspString.match(/localhost/);
       expect(localhostFound).toBeNull();
     });
 
--- a/__tests__/server/middleware/csp.spec.js
+++ b/__tests__/server/middleware/csp.spec.js
@@ -16,6 +16,11 @@
 
 import httpMocks from 'node-mocks-http';
 
+const sanitizeCspString = (cspString) => cspString
+// replaces dynamic ip and nonce to prevent snapshot failures
+  .replace(/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/g, '0.0.0.0')
+  .replace(/nonce-[a-zA-Z0-9]{8}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{12}/, 'nonce-00000000-0000-0000-0000-000000000000');
+
 describe('csp', () => {
   jest.spyOn(console, 'error').mockImplementation(() => {});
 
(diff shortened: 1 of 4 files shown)


stephaniecoates:
None, just additional documentation 

Yes, already complete

to put the logic for adding development domains in One App rather than the root module

No

How was this change tested?
Added unit tests to make sure csp was being written correctly in development

No

No

No

No

No

No

No

No

What is the impact to developers using the app?
Developers will be able to run the app locally and serve modules without having to manually add development domains to their csp

---

No

__tests__/server/middleware/csp.spec.js
```diff
@@ -67,6 +72,39 @@ describe('csp', () => {
...
+      const { updateCSP } = requiredCsp;
+      updateCSP("default-src 'none'; script-src 'self'; connect-src 'self';");
+      cspMiddleware()(req, res, next);
+      // eslint-disable-next-line no-underscore-dangle
+      const headers = res._getHeaders();
+      expect(headers).toHaveProperty('content-security-policy');
+      const cspString = headers['content-security-policy'];
+      const ipFound = cspString.match(/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}:3001/);
+      expect(ipFound).toBeNull();
+      const localhostFound = cspString.match(/localhost:3001/);
```
JAdshead: what about testing for just `localhost` in the snapshot for the test above the port is 5000

__tests__/server/middleware/csp.spec.js
```diff
@@ -67,6 +72,39 @@ describe('csp', () => {
...
+      const requiredCsp = requireCSP();
+      const cspMiddleware = requiredCsp.default;
+      const { updateCSP } = requiredCsp;
+      updateCSP("default-src 'none'; script-src 'self'; connect-src 'self';");
+      cspMiddleware()(req, res, next);
+      // eslint-disable-next-line no-underscore-dangle
+      const headers = res._getHeaders();
+      expect(headers).toHaveProperty('content-security-policy');
+      const cspString = headers['content-security-policy'];
+      const ipFound = cspString.match(/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}:3001/);
```
JAdshead: maybe just test for ip, and ignore the port

src/server/middleware/csp.js
```diff
@@ -58,7 +59,17 @@ export function getCSP() {
 const csp = () => (req, res, next) => {
   const { policy } = cspCache;
   const scriptNonce = uuidV4();
-  const updatedPolicy = insertSource(policy, 'script-src', `'nonce-${scriptNonce}'`);
+  let updatedPolicy;
+  if (process.env.NODE_ENV === 'development') {
+    updatedPolicy = insertSource(
```
JAdshead: #nitpick maybe separate the two `insertSource`s so they are not nested for readability

__tests__/server/middleware/csp.spec.js
stephaniecoates: good catch. This could easily get stale if the `ONE_APP_DEV_CDN_PORT` fallback changes. Will remove and just test for `localhost` and ip

__tests__/server/middleware/csp.spec.js
stephaniecoates: https://github.com/americanexpress/one-app/pull/42#discussion_r384167132

src/server/middleware/csp.js
stephaniecoates: separated in 0cee717

__tests__/server/middleware/csp.spec.js
stephaniecoates: fixed in 0cee717

__tests__/server/middleware/csp.spec.js
stephaniecoates: fixed in 0cee717

PR #46, opened 2020-02-27
docs(client): added documentation for module bundling

nellyk:
This adds documentation on how you can go about bundling your modules, what happens and what files are generated.


docs/api/modules/Bundling-Modules.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+module
+├── README.md
+├── package.json
+├── src
+|   ├── components
+|   |    └── <NameOfComponent.jsx>
+|   |
+|   └── csp.js
```
JamesSingleton: This `csp.js` file is _technically_ only needed in the root module if I remember correctly.

JamesSingleton:
I almost feel like this should go into the generator and not here.

docs/api/modules/Bundling-Modules.md
nellyk: it seems to be generated for any module that is generated using the generator, also this applies root module as well - but i could make it clear

mtomcal: Yes its only for Root modules. We don't have a `--generate-root-module` option for the generator yet but technically only set a CSP for root module.
https://github.com/americanexpress/one-app/blob/master/docs/api/modules/App-Configuration.md#csp

docs/api/modules/Bundling-Modules.md
```diff
@@ -0,0 +1,186 @@
...
+  * [`lint`](#lint-commands)
+  
+## Build Commands
+
+| Command       | Description                                                                                    |
+|---------------|------------------------------------------------------------------------------------------------|
+| `prepare` | Runs `build` before the the module is packed and published. |
+| `prebuild` | Cleans the build directory before a build is done. |
+| `build` | Builds the `one-app` modules.|
+| `watch:build` | Builds the `one-app` modules and watches source files with [`nodemon`](https://nodemon.io/).When source changes, the `one-app` module  runs `npm run build`. *See [`build` Usage](#build-usage).* |
```
JamesSingleton: Missing a space between `.` and `When`

JamesSingleton: Also an extra space between `module` and `runs`

nellyk:
> I almost feel like this should go into the generator and not here.

I'm more inclined to link generator docs to this since this would provide the overall overview of how of how you can bundle modules that you create within the one-app ecosystem

docs/api/modules/Bundling-Modules.md
```diff
@@ -0,0 +1,186 @@
+[👈 Return to Overview](../API.md)
+
+# Bundling
+
+Within `one-app` you can bundle your modules once you've made changes to them. The bundled files are then loaded on the client and server.
+
+## Bundling Modules
```
mtomcal: We may need to move this document as in hindsight to Recipes @nellyk as this definitely lends itself well to a guide.

docs/api/modules/Bundling-Modules.md
```diff
@@ -0,0 +1,186 @@
...
+| `<module-name>.browser.js.map` | Contains the [source map](https://www.html5rocks.com/en/tutorials/developertools/sourcemaps/) which has details about the origin source code. This is loaded on modern browsers to aid in debugging and it is only generated when the `process.env.NODE_ENV` is set to `development`.|
+| `<module-name>.legacy.browser.js` | This build is loaded on legacy browsers. We use [babel-preset-amex](https://github.com/americanexpress/babel-preset-amex) to determine which legacy browsers are supported. Generated using [one-app-bundler](https://github.com/americanexpress/one-app-cli/tree/master/packages/one-app-bundler) which uses [webpack](https://webpack.js.org/) under the hood.|
+| `<module-name>.legacy.browser.js.map` | Contains the [source map](https://www.html5rocks.com/en/tutorials/developertools/sourcemaps/)  which has details about the origin source code. This is loaded on legacy browsers to aid in debugging. it is only generated when the `process.env.NODE_ENV` is set to `development`.|
```
JamesSingleton: Lowercase `it` should be `It` in `It is only generated when`

docs/api/modules/Bundling-Modules.md
```diff
@@ -0,0 +1,186 @@
...
+| `<module-name>.node.js` | This build is loaded on `one-app` . Anything set to be server side rendered would executed once this file is loaded on one-app. |
```
JamesSingleton: There is an extra space after the ` of `one-app` and `.`

docs/api/modules/Bundling-Modules.md
```diff
@@ -0,0 +1,186 @@
...
+| `<module-name>.node.js.map` | Contains the [source map](https://www.html5rocks.com/en/tutorials/developertools/sourcemaps/)  which has details about the origin source code. This allows you to debug code running on one-app. This is only generated when the `process.env.NODE_ENV` is set to `development`. |
```
JamesSingleton: There is an extra space between `/sourcemaps/)` and `which`

docs/api/modules/Bundling-Modules.md
anescobar1991: And we now do have that option in the generator since https://github.com/americanexpress/one-app-cli/pull/25 was merged!

nellyk: Thank you, i've removed this.

docs/api/modules/Bundling-Modules.md
nellyk: thanks for catching this updated c5b191e

docs/api/modules/Bundling-Modules.md
nellyk: Agreed, i've moved this to recipes updated c5b191e

docs/api/modules/Bundling-Modules.md
nellyk: thanks for catching this updated c5b191e

docs/api/modules/Bundling-Modules.md
nellyk: Thanks for catching this updated c5b191e

docs/recipes/Bundling-Modules.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+module
+├── README.md
+├── package.json
+├── src
+|   ├── components
+|   |    └── <NameOfComponent.jsx>
+|   |
+|   └── csp.js
```
infoxicator: `csp.js` is missing on the previous structure before line 25, which might imply that the `npm run build` command generates that file. We should remove it if considered irrelevant for this example or keep it as it is the default file structure for a freshly created module using generator...

nellyk: thank you for catching this, i've updated this 6dd6088

PR #48, opened 2020-02-28
fix(devCdn): proxy modules through dev cdn

anescobar1991:
See https://github.com/americanexpress/one-app-dev-cdn/pull/4

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -158,9 +158,9 @@
       }
     },
     "@americanexpress/one-app-dev-cdn": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/@americanexpress/one-app-dev-cdn/-/one-app-dev-cdn-3.0.0.tgz",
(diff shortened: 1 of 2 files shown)


anescobar1991:
**What is the communication plan for this change?**
One App Dev CDN docs are updated to discuss the changes
**Does any documentation need to be updated or added to account for this change? If so was it done already?**
n/a
**What is the motivation for this change?**
See americanexpress/one-app-dev-cdn#4
**Should these changes also be applied to a maintenance branch or any other long lived branch?**
n/a
**How was this change tested?**
ran locally with both a remote module map and without one, also with and without proxy env vars
**Does this change require cross browser checks? Why or why not?**
no
**Does this change require a performance test prior to merging? Why or why not?**
no as this is a dev only change
**Could this be considered a breaking change? Why or why not?**
no
**Does the change impact caching?**
no
**Does the change impact HTTP headers?**
no
**Does the change have any new infrastructure requirements?**
no
**Does the change affect other versions of the app?**
no
**Does the change require additional environment variables?**
no
**What is the impact to developers using the app?**
fixes a bug for local development and allows those inside corporate firewalls to use module maps outside their firewall by use of the proxy env vars
**What is the change to the size of assets?**
n/a
**Should integration tests be added to protect against future regressions on this change?**
n/a

PR #54, opened 2020-03-04
Bug fixes for mocking data

enuchi:
## Description
- FIx csp directives to include dev proxy server port.
- Fix path joining in state config.

## Motivation and Context
Solves csp directives not including dev proxy server port.
Solves incorrect double slash `//` in path naming.  

## How Has This Been Tested?
Includes tests to validate desired behavior.

## Types of Changes

## Checklist:
- [x] This change impacts HTTP headers.

Dev proxy server port should appear on csp directives, allowing proper testing with parrot.
Allows defining paths in mock endpoints with leading or no leading slash.

### DIFF
--- a/src/server/middleware/csp.js
+++ b/src/server/middleware/csp.js
@@ -65,7 +65,12 @@ const csp = () => (req, res, next) => {
     updatedPolicy = insertSource(
       updatedScriptSrc,
       'connect-src',
-      `${ip.address()}:${process.env.HTTP_ONE_APP_DEV_CDN_PORT} localhost:${process.env.HTTP_ONE_APP_DEV_CDN_PORT}`
+      [
+        `${ip.address()}:${process.env.HTTP_ONE_APP_DEV_CDN_PORT}`,
+        `localhost:${process.env.HTTP_ONE_APP_DEV_CDN_PORT}`,
+        `${ip.address()}:${process.env.HTTP_ONE_APP_DEV_PROXY_SERVER_PORT}`,
+        `localhost:${process.env.HTTP_ONE_APP_DEV_PROXY_SERVER_PORT}`,
+      ].join(' ')
     );
   } else {
     updatedPolicy = insertSource(policy, 'script-src', `'nonce-${scriptNonce}'`);
--- a/src/server/utils/stateConfig.js
+++ b/src/server/utils/stateConfig.js
@@ -17,6 +17,7 @@
 import ip from 'ip';
 import fs from 'fs';
 import path from 'path';
+import url from 'url';
 import envVarAllowList from './envVarAllowList';
 import snakeCaseToCamelCase from './snakeCaseToCamelCase';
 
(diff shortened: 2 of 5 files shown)


src/server/utils/stateConfig.js
```diff
@@ -76,7 +77,12 @@ if (process.env.NODE_ENV === 'development' && fs.existsSync(pathToDevEndpoints))
   // eslint-disable-next-line global-require,import/no-dynamic-require
   const devEndpoints = require(pathToDevEndpoints)();
   Object.entries(devEndpoints).forEach(([configName, { devProxyPath }]) => {
-    const value = `http://${ipAddress}:${SERVICES_PORT}/${devProxyPath}`;
+    const value = url.format({
```
JamesSingleton: What is beneficial of making this change compared to what is already there as they result in the same thing.

src/server/middleware/csp.js
```diff
@@ -65,7 +65,12 @@ const csp = () => (req, res, next) => {
     updatedPolicy = insertSource(
       updatedScriptSrc,
       'connect-src',
-      `${ip.address()}:${process.env.HTTP_ONE_APP_DEV_CDN_PORT} localhost:${process.env.HTTP_ONE_APP_DEV_CDN_PORT}`
+      [
+        `${ip.address()}:${process.env.HTTP_ONE_APP_DEV_CDN_PORT}`,
+        `localhost:${process.env.HTTP_ONE_APP_DEV_CDN_PORT}`,
+        `${ip.address()}:${process.env.HTTP_ONE_APP_DEV_PROXY_SERVER_PORT}`,
+        `localhost:${process.env.HTTP_ONE_APP_DEV_PROXY_SERVER_PORT}`,
```
JAdshead: maybe we should not be so opinionated on the ports during development.
@anescobar1991 thoughts on :
```suggestion
        `${ip.address()}:*`,
        'localhost:*',
```

anescobar1991: As long as its just for development I dont see any issues. 👍

anescobar1991: Although maybe this would confuse users who dont realize that `localhost` is not just open and when they get to production they would expect it to work but it doesnt?

enuchi:
Sorry, had to rewrite commits to add email address so I could add signed CLA. @JamesSingleton can you pull in master again if needed?

src/server/middleware/csp.js
JAdshead: i think that might be the case anyway by opening up some ports

JAdshead: @enuchi would you be able to make this change and then i think this is good to go.

anescobar1991: but those ports are reserved for one app usage by nature of them being env vars used by one app though no?

JAdshead: thats a fair point, so maybe leave as is and open it up later if we find a need

anescobar1991: yeah! if we do the opposite it becomes a breaking change!

JamesSingleton:
Here is the bundle size report:

### 📊 Bundle Size Report
| file name             | size on disk |    gzip |
| :-------------------- | -----------: | ------: |
| app.js                |      112.7KB |  31.4KB |
| runtime.js            |         15KB |   5.3KB |
| vendors.js            |      128.1KB |  37.9KB |
| app~vendors.js        |      403.5KB | 105.2KB |
| legacy/app.js         |      119.4KB |    33KB |
| legacy/runtime.js     |         15KB |   5.3KB |
| legacy/vendors.js     |        163KB |  44.8KB |
| legacy/app~vendors.js |      409.8KB | 106.9KB |

PR #56, opened 2020-03-06
chore(logging): remove start() on oneAppDevCdnSpinner

JamesSingleton:
## Description

This fixes watchLocalModules.spec.js silently failing when you run the test individually. Furthermore, it is not enabled on like 97.

## Motivation and Context

It fixes a unit test that was silently failing. If you run the unit tests all together it will pass. However, if you run this test by itself, it will fail.

## Types of Changes

## Checklist:


src/server/utils/logging/development-formatters/friendly.js
```diff
@@ -36,7 +36,7 @@ function createSpinner(opts) {
 const oneAppDevCdnSpinner = createSpinner({
   text: 'Starting one-app-dev-cdn server...',
   spinner: 'bouncingBar',
```
Francois-Esquire: Would you need to add the option `isEnabled: false`?

JamesSingleton: I am not sure, would like to get input from @10xLaCroixDrinker or @mtomcal on this part

PR #58, opened 2020-03-11
docs(styling): add styling recipe docs

JamesSingleton:
## Description

Adds documentation for adding styles within One App

## Motivation and Context

Adds needed documentation

## How Has This Been Tested?

N/A as this is documentation.

## Types of Changes

## Checklist:


infoxicator:
I know it is not recommended but is it worth including how to import global css by wrapping styles in `:global {}` thoughts?

docs/recipes/Adding-Styles.md
```diff
@@ -2,6 +2,52 @@
...
+In order to access styles within your application, you will need to add the stylesheet to the `<head>`. You can achieve this by using [react-helmet](https://github.com/nfl/react-helmet).
+
+```jsx
+import React, { Fragment } from 'react';
+import { Helmet } from 'react-helmet';
+
+const MyRoot = ({ children }) => (
+  <Fragment>
+    <Helmet>
+      <link rel="stylesheet" href="https://www.example.com/styles.min.css" />
```
infoxicator: @JamesSingleton Maybe include a reminder that the domain has to be added to the `csp`? or is it implied?

docs/recipes/Adding-Styles.md
```diff
@@ -2,6 +2,52 @@
...
+### Example Class Names
+
+```css
+/* good */
+.myClass {
+  color: red;
+}
+
+/* bad, won't work */
+.my-class {
```
infoxicator: Technically this works if you use [bracket notation] instead of .dot notation to include the styles, but I agree it is not a recommended pattern.

docs/recipes/Adding-Styles.md
```diff
@@ -2,6 +2,52 @@
 
 # Adding Styles
 
-Recipe is forthcoming.
+## Adding a CSS File to Root Module
```
infoxicator: global styles can also be added to normal modules as well in the Helmet of that module, I don't know if there is a benefit of doing that and might keep modules independent? Maybe mention that it is also a possibility?

JamesSingleton: I mean yes it is possible. However, I don't think that would be the recommended route. For example, lets say 5 modules add the same CSS file or slightly different CSS files that override each other.

docs/recipes/Adding-Styles.md
JamesSingleton: I can mention that.

docs/recipes/Adding-Styles.md
infoxicator: I am having a hard time thinking what's the best approach in this scenario, because I could also argue that I don't want to load the styles in the root module because it will make the module dependent and I think Helmet will resolve duplications fine... on the other hand as you said that could cause collisions... in summary **I don't think adding global css to the root module is a good idea either**. The best option is to use css modules at the module level. I will work on a POC loading bulma for example from the module itself instead of CDN and then import another library i.e SemanticUI to another module and hopefully there shouldn't be any collisions.

JamesSingleton:
> I know it is not recommended but is it worth including how to import global css by wrapping styles in `:global {}` thoughts?

@infoxicator this is already mentioned in the CSS Modules docs https://github.com/css-modules/css-modules, so I don't think we need to mention it here.

docs/recipes/Adding-Styles.md
JamesSingleton: So my _concern_ with that is lets say one module loads a completely different style sheet than another but are used within the same application? They will most likely conflict with each other.

mtomcal: Almost better to leave unstyled a reusable child Holocron Module. Root Modules are intended to be tenancy specific but child Modules ideally should be reusable. Therefore, Child Modules would need to be fairly generic style-wise but the Tenancy Root contains the overall theme via Helmet. Yeah I'm not a fan of having Child Modules automatically dependent on a specific  Root Module provider. If there's a way for the Child Module to operate alone without a Provider or an optional lightweight provider locally if context doesn't exist at the Root level, the better.

jsmartfo: I agree with the commentary above about independence but this is about recipes, what is possible and the trade-offs. 

We have an opinion on how we would do styles (backpacked at module level) but not every user of one-app will have the same use cases. 

There needs to be some form of table to support this. 

1. - Stylesheets loaded at root module (global for app, all modules upgrade cycle the same)
2. - Stylesheets loaded at child module level (collisions can occur, multiple loads of same styles)
3. - Insulated at module level using CSS Modules (prone to some bloat, possibility of less re-use)
4. - Insulated at Component library level using CSS Modules or other JS pattern (some bloat, entirely insular, re-use achieved)

There are many ways to skin this. Our proposed pattern has been option 4, other users may not have the same use cases. Let's just explain the differences somehow.

infoxicator:
> @infoxicator this is already mentioned in the CSS Modules docs https://github.com/css-modules/css-modules, so I don't think we need to mention it here.

... I feel like even though is repetitive, that would be a question we will get a lot, is a link to css modules enough? otherwise I volunteer to answer the question in StackOverflow 😄

docs/recipes/Adding-Styles.md
infoxicator: I agree with @jsmartfo we should advise what's the recommended way but also mention the pros/cons of the alternatives... We don't want people using incorrect patterns but also we don't want them complaining that one app is restrictive/prescriptive

docs/recipes/Adding-Styles.md
```diff
@@ -2,6 +2,157 @@
...
 # Adding Styles
 
-Recipe is forthcoming.
+## CSS Modules
 
-[☝️ Return To Top](#adding-styles)
\ No newline at end of file
+Since CSS is global, One App uses the [CSS Modules](https://github.com/css-modules/css-modules) pattern to avoid class name collisions. This pattern should not be too different than what you are already used to. The main differences will be the strict use of camelCased class names and the way you import the class names.
+
+### Example Class Names
```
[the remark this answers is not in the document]
JamesSingleton: The examples are right below this header though.

docs/recipes/Adding-Styles.md
Francois-Esquire: Agree here, the wording could be softer as this is valid `css`, however incompatible currently.

docs/recipes/Adding-Styles.md
```diff
@@ -2,6 +2,171 @@
...
+* [Using Reakit with One App](#using-reakit-with-one-app)
+  * [Add Reakit to Root Module](#add-reakit-to-root-module)
+  * [Using Reakit in a Child Module](#using-reakit-in-a-child-module)
+  * [Share Reakit Across Your Modules](#share-reakit-across-your-modules)
+    * [Add Reakit as a `providedExternals` in your Root Module](#add-reakit-as-a-providedexternals-in-your-root-module)
+    * [Add Reakit as a `requiredExternals` in your Child Module](#add-reakit-as-a-requiredexternals-in-your-child-module)
+
+## CSS Modules
+
+Since CSS is global, One App uses the [CSS Modules](https://github.com/css-modules/css-modules) pattern to avoid class name collisions. The main differences will be the strict use of camelCased class names and the way you import the class names.
```
Francois-Esquire: Could we break this up to 100 chars per line, and anywhere else in the doc please

docs/recipes/Adding-Styles.md
```diff
@@ -2,6 +2,181 @@
...
+
+### Example Class Names
+
+```css
+/* good */
+.myClass {
+  color: red;
+}
+
+/* bad, won't work */
```
Francois-Esquire:
```suggestion
/* not currently supported due to purge-css */
```

PR #63, opened 2020-03-13
fix(stateConfig): use localhost rather than ip for dev

anescobar1991:
In order to start using One App Runner we need One App to make calls to the dev-proxy server using `localhost` rather than the resolved IP address. The reason is that when running One App as a docker container the IP address inside that container is not resolvable by the host.

## Description

Updated the state config modifications we make when there is a dev.endpoints.js file to use localhost rather than the IP address

## Motivation and Context

In order to start using One App Runner we need One App to make calls to the dev-proxy server using `localhost` rather than the resolved IP address. The reason is that when running One App as a docker container the IP address inside that container is not resolvable by the host.

## How Has This Been Tested?

Unit tests were updated and made sure to run locally with parrot and a dev endpoints file.

## Types of Changes

## Checklist:

Allows them to use one app runner

### DIFF
--- a/src/server/utils/stateConfig.js
+++ b/src/server/utils/stateConfig.js
@@ -14,7 +14,6 @@
  * permissions and limitations under the License.
  */
 
-import ip from 'ip';
 import fs from 'fs';
 import path from 'path';
 import url from 'url';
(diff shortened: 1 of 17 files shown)


src/server/utils/stateConfig.js
```diff
@@ -79,7 +79,7 @@ if (process.env.NODE_ENV === 'development' && fs.existsSync(pathToDevEndpoints))
   Object.entries(devEndpoints).forEach(([configName, { devProxyPath }]) => {
     const value = url.format({
       protocol: 'http',
-      hostname: ipAddress,
+      hostname: 'localhost',
```
JamesSingleton: If we are going to remove `ipAddress` can you also remove https://github.com/americanexpress/one-app/blob/master/src/server/utils/stateConfig.js#L75?

Francois-Esquire: Also the import please:

https://github.com/americanexpress/one-app/blob/master/src/server/utils/stateConfig.js#L17

PR #81, opened 2020-03-26
feat(runtime): add csp reporting url env var

JAdshead:
## Description

Adding additional env var allow the configuration of the csp-violations report url
## Motivation and Context

Currently only a single environment variable is used to configuring reporting urls which has lead to misconfiguration.

## How Has This Been Tested?

Unit & Integrations tests

## Types of Changes

## Checklist:

Allows configuration of csp-violation and error reporting urls


JamesSingleton:
@JAdshead,

Any reason why there are changes in `package-lock.json` when there are none in the `package.json`?

JAdshead:
> ONE_CLIENT_CSP_REPORTING_URL

JAdshead:
> Do we need to update the route address for the middleware as well for reporting url?
> 
> https://github.com/americanexpress/one-app/blob/master/src/server/ssrServer.js#L69
> 
> Not sure if there is a reason for it or not. My only concern if I set `ONE_CLIENT_CSP_REPORTING_URL` to something different like `/csp-violation` instead, it would break with the hard coded route.. We also prob don't need that middleware if we have a complete url with `https://my-csp-reporter.com/violation`.

@Francois-Esquire one-app provides the reporting endpoints to allow users to optionally report back and log on the one-app server, there are cases where they will want to standup their own server for handling reports. setting ONE_CLIENT_CSP_REPORTING_URL and ONE_CLIENT_REPORTING_URL allows dependencies to consume those values and not be coupled to one-app's api.

docs/api/server/Environment-Variables.md
```diff
@@ -421,11 +422,41 @@ ONE_CLIENT_REPORTING_URL=https://my-app-errors.com/client
...
 
+## `ONE_CLIENT_CSP_REPORTING_URL`
+
+⚠️ Required In Production
+
+**Runs In**
+* ✅ Production
+* ✅ Development
+
+URL where browser should send client side errors to. Defaults to `/_/report/security/csp-violation` if `NODE_ENV` is set to
```
10xLaCroixDrinker:
```suggestion
URL where browser should send [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) errors to. Defaults to `/_/report/security/csp-violation` if `NODE_ENV` is set to
```

docs/api/server/Environment-Variables.md
```diff
@@ -421,11 +422,41 @@ ONE_CLIENT_REPORTING_URL=https://my-app-errors.com/client
...
+ONE_CLIENT_CSP_REPORTING_URL=https://my-app-csp-violations.example.com/client
+```
+
+**Default Value**
+```bash
+# if NODE_ENV=development
+ONE_CLIENT_REPORTING_URL=/_/report/security/csp-violation
+# else
+ONE_CLIENT_REPORTING_URL=undefined
+```
```
10xLaCroixDrinker:
```suggestion
ONE_CLIENT_CSP_REPORTING_URL=/_/report/security/csp-violation
# else
ONE_CLIENT_CSP_REPORTING_URL=undefined
```
```

File: docs/api/server/Environment-Variables.md [hunk 3]
```diff
@@ -421,11 +422,41 @@ ONE_CLIENT_REPORTING_URL=https://my-app-errors.com/client
 **Default Value**
 ```bash
 # if NODE_ENV=development
-ONE_CLIENT_REPORTING_URL=/_
+ONE_CLIENT_REPORTING_URL=/_/report/errors
```
10xLaCroixDrinker: If you're going to do this, may as well change the env var name altogether.
```suggestion
ONE_CLIENT_ERROR_REPORTING_URL=/_/report/errors
```

10xLaCroixDrinker: Won't this also require a change to one-app-ducks?

JAdshead: changing the name would require a change which is why i left it, currently one-app-ducks is treating this as `/_/report/errors` - https://github.com/americanexpress/one-app-ducks/blob/master/src/errorReporting.js#L109

10xLaCroixDrinker: carry on 🖖

prod-sample/one-app/base.env
```diff
@@ -1,5 +1,6 @@
 HOLOCRON_MODULE_MAP_URL=https://sample-cdn.frank/module-map.json
-ONE_CLIENT_REPORTING_URL=https://one-app:8443/_
+ONE_CLIENT_REPORTING_URL=https://one-app:8443/_/report/errors
```
JAdshead, line 2: @10xLaCroixDrinker ☝️

JAdshead:
> This should be updated as well: https://github.com/americanexpress/one-app/blob/master/prod-sample/one-app/base.env#L2
> 
> Can you validate that this is covered everywhere?

I have been through and checked other repos for use of `ONE_CLIENT_REPORTING_URL/reportingUrl` and only occurrences i could find were in the generator and one-app-ducks. The generator is the only place that requires an update

mtomcal:
Do we have documented any how errors are POST-ed to the `ONE_CLIENT_CSP_REPORTING_URL`?

JAdshead:
> Do we have documented anywhere how errors are POST-ed to the `ONE_CLIENT_CSP_REPORTING_URL`?

This would be elaborated on in a creating CSP recipe that I don't think exists

PR #91, opened 2020-04-03
Feat/integration test variables

infoxicator:
Add Heroku and Surge environment variables
## Description
Added environment variables so one app developers can configure their integration test environment in Heroku and surge so they can run the integration tests in their own deployed application and catch any integration errors that might occur before creating a pull request to master.

## Motivation and Context

Improves resilience and allows one app developers to run integration tests against a deployed instance of one app 

## How Has This Been Tested?

Tested on my One App heroku instance here 
https://one-app-integration-test.herokuapp.com/success

## Types of Changes

## Checklist:


scripts/run-integration-tests.travis.sh
```diff
@@ -2,15 +2,15 @@
 set -ev
 docker build -t one-app:at-test .
 if [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "master" ]; then
-  npm run build:sample-modules -- --archive-built-artifacts --bundle-statics-origin=https://one-app-statics.surge.sh
-  echo 'https://one-app-sample.herokuapp.com' >> sample-module-bundles/CORS && \
-  npx surge teardown one-app-statics.surge.sh && \
-  npx surge sample-module-bundles one-app-statics.surge.sh && \
+  npm run build:sample-modules -- --archive-built-artifacts --bundle-statics-origin=$SURGE_DOMAIN
```
JamesSingleton, line 5: Can we make these more _generic_? If in the future we go away from surge/heroku we won't have to rename these.

infoxicator: Unfortunately, the script is very specific on how surge and Heroku deployments work, if we need to use a different deployment system, it will likely require different variables / not require them at all

PR #101, opened 2020-04-09
feat(fetch): switch to cross-fetch instead of isomorphic-fetch

JAdshead:
## Description

Switching to [cross-fetch](https://github.com/lquixada/cross-fetch) instead of [isomorphic-fetch](https://github.com/matthew-andrews/isomorphic-fetch) as the later has not been maintained for 5 years.

## Motivation and Context

isomorphic-fetch as the later has not been maintained for 5 years and provides outdated versions of both node-fetch and WHATWG fetch

## How Has This Been Tested?

Unit and integration tests.

## Types of Changes

## Checklist:

Provides a fetch that is closer to the spec.


infoxicator:
@JAdshead Integration tests pass so looks like everything is working same as with `isomorphic-fetch` 😁, if I want to run some manual tests, is this the only dep I need? "@americanexpress/one-app-bundler": "^6.4.0",

__tests__/integration/one-app.spec.js
```diff
@@ -876,8 +879,8 @@ describe('Tests that can run against either local Docker setup or remote One App
           }
         );
         expect(response.status).toBe(200);
-        expect(response.headers._headers).toHaveProperty('access-control-allow-origin');
-        expect(response.headers._headers['access-control-allow-origin']).toEqual(['test.example.com']);
+        expect(response.headers.raw()).toHaveProperty('access-control-allow-origin');
```
mtomcal, line 882: Would this new API of `raw` affect other wrappers or tools?

JAdshead:
> @JAdshead Integration tests pass so looks like everything is working same as with `isomorphic-fetch` 😁, if I want to run some manual tests, is this the only dep I need? "@americanexpress/one-app-bundler": "^6.4.0",

Yes you will need the latest bundler

__tests__/integration/one-app.spec.js
JAdshead: Should not, the versions of `node-fetch` and `whatwg-fetch` are much closer aligned. the `headers._headers` is a node-fetch@v1 only api.

mtomcal:
This looks good and given how many Integration tests where we use `fetch` this gives me confidence this is a non breaking swapout.

PR #111, opened 2020-04-21
feat(createRequestHtmlFragment): implemented circuit breaker

10xLaCroixDrinker:
## Description

This implements a circuit breaker using opossum. The circuit opens when event loop delay is > 250ms (by default- is configurable). When the circuit is open there is no SSR. The circuit will go into a half open state after 10s. At that point it will attempt the full request again, the circuit will close.

## Motivation and Context

This provides significant performance improvements.

## How Has This Been Tested?

In addition to the unit tests, I did some load testing locally. I ran a json server using `json-server http://jsonplaceholder.typicode.com/db -d 800 -p 1337`, then changed SSR Frank to make its requests to `http://localhost:1337/todos`. Once that was set up I called autocannon with `autocannon localhost:3000/healthy-frank/ssr-frank -c 200 -d 30 --headers "Accept-Language= en-USen;q=0.5" --headers "User-Agent= curl/7.5.0"
`. Below are the results for master and for this branch.

(Note the circuit breaker is disabled for perf tests by setting `eventLoopDelayThreshold` to `Infinity`. To validate locally, that line should be removed.

### master

```
Running 30s test @ http://localhost:3000/healthy-frank/ssr-frank
200 connections

┌─────────┬────────┬─────────┬─────────┬─────────┬────────────┬─────────┬────────────┐
│ Stat    │ 2.5%   │ 50%     │ 97.5%   │ 99%     │ Avg        │ Stdev   │ Max        │
├─────────┼────────┼─────────┼─────────┼─────────┼────────────┼─────────┼────────────┤
│ Latency │ 980 ms │ 1110 ms │ 1426 ms │ 1517 ms │ 1124.38 ms │ 99.5 ms │ 1590.64 ms │
└─────────┴────────┴─────────┴─────────┴─────────┴────────────┴─────────┴────────────┘
┌───────────┬─────┬──────┬─────────┬─────────┬─────────┬─────────┬─────────┐
│ Stat      │ 1%  │ 2.5% │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min     │
├───────────┼─────┼──────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Req/Sec   │ 0   │ 0    │ 184     │ 200     │ 175.1   │ 36.24   │ 144     │
├───────────┼─────┼──────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Bytes/Sec │ 0 B │ 0 B  │ 11.1 MB │ 13.4 MB │ 9.75 MB │ 3.82 MB │ 1.73 MB │
└───────────┴─────┴──────┴─────────┴─────────┴─────────┴─────────┴─────────┘

Req/Bytes counts sampled once per second.

5k requests in 30.14s, 292 MB read
```

### This branch

```
Running 30s test @ http://localhost:3000/healthy-frank/ssr-frank
200 connections

┌─────────┬────────┬────────┬────────┬────────┬──────────┬──────────┬────────────┐
│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg      │ Stdev    │ Max        │
├─────────┼────────┼────────┼────────┼────────┼──────────┼──────────┼────────────┤
│ Latency │ 239 ms │ 291 ms │ 441 ms │ 463 ms │ 312.1 ms │ 99.35 ms │ 1450.22 ms │
└─────────┴────────┴────────┴────────┴────────┴──────────┴──────────┴────────────┘
┌───────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┬─────────┐
│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%   │ Avg     │ Stdev   │ Min     │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Req/Sec   │ 229     │ 229     │ 661     │ 807     │ 637.5   │ 124.65  │ 229     │
├───────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Bytes/Sec │ 1.91 MB │ 1.91 MB │ 5.51 MB │ 6.72 MB │ 5.31 MB │ 1.03 MB │ 1.91 MB │
└───────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┴─────────┘

Req/Bytes counts sampled once per second.

19k requests in 30.09s, 159 MB read
```

## Types of Changes

## Checklist:

- [x] Performance tests should be ran against the server prior to merging.

No impact to developers other than SSR turning off in the cases that the circuit opens


docs/api/modules/App-Configuration.md
```diff
@@ -267,6 +267,26 @@ if (!global.BROWSER) {
...
+```js
+if (!global.BROWSER) {
+  Module.appConfig = {
+    eventLoopLagThreshold: Number,
+  };
+}
+```
+
+The `eventLoopLagThreshold` directive accepts a number representing the threshold of the event loop lag (in milliseconds) before opening the circuit in our circuit breaker created with [Opossum](https://nodeshift.dev/opossum/). Once the circuit is open, it will remain open for 10 seconds and close at that time pending the event loop lag. The default value is `30`. If you desire to disable the circuit completely, set this value to `Infinity`.
+
```
mtomcal, line 285: Maybe add some text like:

```In practice, `eventLoopLagThreshold` allows for tuning server side rendering (SSR) of Modules. We may increase request throughput by temporarily disabling SSR at high load through [Event Loop Lag](some-link-to-article-on-event-loop-lag) monitoring.```

File: __tests__/server/middleware/createRequestHtmlFragment.spec.js [hunk 2]
```diff
@@ -229,4 +245,86 @@ describe('createRequestHtmlFragment', () => {
     expect(next).toHaveBeenCalled();
     /* eslint-enable no-console */
   });
+
+  it('should open the circuit when event loop lag is > 30ms', async () => {
+    expect.assertions(5);
```
JAdshead, line 250: is this required with async test ?

__tests__/server/utils/circuitBreaker.spec.js
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+import CircuitBreaker from 'opossum';
+import breaker, {
+  setEventLoopLagThreshold,
+  getEventLoopLagThreshold,
+} from '../../../src/server/utils/circuitBreaker';
+
+describe('Circuit breaker', () => {
+  it('should be an opossum circuit breaker', () => {
```
JAdshead, line 24: does this matter if everything else works ?

prod-sample/sample-modules/frank-lloyd-root/0.0.0/src/config.js
```diff
@@ -18,6 +18,7 @@ import csp from './csp';
 import createFrankLikeFetch from './createFrankLikeFetch';
 
 export default {
+  eventLoopLagThreshold: Infinity,
```
JAdshead, line 21: do these not work with the default ?

10xLaCroixDrinker: This prevents the circuit from opening during the integration tests

__tests__/server/middleware/createRequestHtmlFragment.spec.js
10xLaCroixDrinker: yes

__tests__/server/utils/circuitBreaker.spec.js
10xLaCroixDrinker: not necessarily

__tests__/server/middleware/createRequestHtmlFragment.spec.js
JAdshead: checked locally, `expect.assertions(5);` can be removed.

__tests__/server/middleware/createRequestHtmlFragment.spec.js
```diff
@@ -14,24 +14,35 @@
...
 const renderForStringSpy = jest.spyOn(reactRendering, 'renderForString');
 const renderForStaticMarkupSpy = jest.spyOn(reactRendering, 'renderForStaticMarkup');
+const msToNs = (n) => n * 1e6;
 
 describe('createRequestHtmlFragment', () => {
   jest.spyOn(console, 'error').mockImplementation(() => {});
+  const realHrtime = process.hrtime;
+  const mockHrtime = (...args) => realHrtime(...args);
+  mockHrtime.bigint = jest.fn();
+  process.hrtime = mockHrtime;
```
mcollina, line 45: I think this should not really required here, e.g. the breaker should be automatically disabled in tests.

src/server/utils/circuitBreaker.js
```diff
@@ -0,0 +1,66 @@
...
+const options = {
+  timeout: false, // Do not use a timeout
+  errorThresholdPercentage: 1, // When 1% of requests fail, open the circuit
+  resetTimeout: 10e3, // After 10 seconds, try again.
+};
+
+const breaker = new CircuitBreaker(getModuleData, options);
+// Just need to connect opossum to prometheus
+// eslint-disable-next-line no-unused-vars
+const metrics = new PrometheusMetrics(breaker, register);
```
mcollina, line 50: I would probably not expose  this at this level. Why was it added here?

src/server/utils/circuitBreaker.js
```diff
@@ -0,0 +1,66 @@
...
+  await immediate();
+  const end = process.hrtime.bigint();
+  const diff = (end - start) / BigInt(1e6);
+  // Open the circuit if event loop lag is greater than threshold
+  if (diff > eventLoopLagThreshold) {
+    throw new Error(`Opening circuit, event loop lag (${diff}ms) is > eventLoopLagThreshold (${eventLoopLagThreshold}ms)`);
+  }
+}, 100);
+
+export default breaker;
```
mcollina, line 66: I think a more reusable and testable implementation would be to make this a factory, so we can avoid pulling in holocron in  here, but rather have the `getModuleData` passed in to the factory instead.

__tests__/server/middleware/createRequestHtmlFragment.spec.js
10xLaCroixDrinker: This is just needed for the tests that are asserting on the functionality of the breaker

src/server/utils/circuitBreaker.js
10xLaCroixDrinker: Just keeping it close to the code, but yes it makes more since to go in [`src/server/metrics`](https://github.com/americanexpress/one-app/tree/master/src/server/metrics)

__tests__/server/middleware/createRequestHtmlFragment.spec.js
10xLaCroixDrinker: expect.assertions ensures we get passed the promise to all the assertions

src/server/utils/circuitBreaker.js
```diff
@@ -0,0 +1,66 @@
...
+import { register } from 'prom-client';
+import { composeModules, getModule } from 'holocron';
+import { getServerStateConfig } from './stateConfig';
+
+const immediate = promisify(setImmediate);
+const { rootModuleName } = getServerStateConfig();
+let eventLoopLagThreshold = 30;
+
+export const setEventLoopLagThreshold = (n) => {
+  eventLoopLagThreshold = Number(n) || 30;
```
10xLaCroixDrinker, line 31: @mcollina what are your thoughts on this default value?

mcollina: 30ms is _very_ aggressive. I would increase quite a bit. The 100-500ms range seems more ok for an SSR app (most React rendering go up to 80-100ms).

src/server/utils/circuitBreaker.js
```diff
@@ -0,0 +1,66 @@
...
+// Just need to connect opossum to prometheus
+// eslint-disable-next-line no-unused-vars
+const metrics = new PrometheusMetrics(breaker, register);
+
+breaker.fallback(() => true);
+
+breaker.healthCheck(async () => {
+  if (!getModule(rootModuleName)) return;
+  const start = process.hrtime.bigint();
+  await immediate();
```
mcollina, line 57: You might want to use https://nodejs.org/dist/latest-v12.x/docs/api/perf_hooks.html#perf_hooks_perf_hooks_monitoreventloopdelay_options to measure event loop latency. It adds less overhead.

10xLaCroixDrinker:
@mcollina I did some refactoring here, think you could review again?

prod-sample/sample-modules/frank-lloyd-root/0.0.0/src/config.js
```diff
@@ -18,6 +18,7 @@ import csp from './csp';
 import createFrankLikeFetch from './createFrankLikeFetch';
 
 export default {
+  eventLoopDelayThreshold: Infinity,
```
mcollina, line 21: Should this be set to a smaller amount?

10xLaCroixDrinker: This is just to prevent the circuit from opening during the integration tests

src/server/middleware/createRequestHtmlFragment.jsx
```diff
@@ -69,11 +77,22 @@ export default function createRequestHtmlFragment({ createRoutes }) {
...
-      await dispatch(composeModules(routeModules));
-
       const state = store.getState();
       const disableScripts = state.getIn(['rendering', 'disableScripts']);
       const renderPartialOnly = state.getIn(['rendering', 'renderPartialOnly']);
+
+      if (disableScripts || renderPartialOnly) {
+        await dispatch(composeModules(routeModules));
+      } else {
+        const fallback = await breaker.fire({ dispatch, modules: routeModules });
```
JAdshead, line 87: maybe rename breaker to show its related to getModuleData

```suggestion
        const fallback = await getModuleDataBreaker.fire({ dispatch, modules: routeModules });
```

PR #119, opened 2020-04-27
feat(pwa/web-manifest): add middleware & support config

Francois-Esquire:
Add Web App Manifest PWA feature for `one-app`; configurable by `appConfig` and delivers the
addition of `webManifest` key as the manifest body.

## Description

This PR extends support for the `manifest.webmanifest`, created from object and added to the HTML document during SSR. It can be removed when explicitly passing `null` or `false` for the `webManifest` key under the `appConfig.pwa` root configuration.

## Motivation and Context

Addition of PWA technology for `one-app`. It allows the client side to be installable on devices with and provides meta-data on the application.

## How Has This Been Tested?

Yes, both unit tests and integration tests.

## Types of Changes

## Checklist:

It gives the option to opt in to using the web app manifest in One App.


prod-sample/sample-modules/frank-lloyd-root/0.0.3/src/pwa.js
```diff
@@ -14,7 +14,46 @@
  * permissions and limitations under the License.
  */
 
+export const defaultSplashScreen = 'http://localhost:3001/static/modules/frank-lloyd-root/0.0.3/assets/pwa-icon-512px.png';
```
JAdshead, line 17: this might cause some issues when running the remote integration tests

docs/api/modules/App-Configuration.md
```diff
@@ -204,13 +208,42 @@ if (!global.BROWSER) {
...
       scope: '/',
+      // the web app manifest can be directly incorporated in the PWA config
+      webManifest: {
+        // the full name is the official name of a given PWA
+        name: 'My App Name',
+        // the short name is used by mobile devices to label your home screen icon
+        short_name: 'My App',
+        // the description is a good piece of meta-data to include for a short description
+        // which can be used with presenting your PWA
+        description: 'My PWA app.',
```
nellyk, line 219: Can you use all the features listed here https://developer.mozilla.org/en-US/docs/Web/Manifest or on those listed here?
 From the article i noticed that `icons` is mandatory is mandatory. Do we have some features that are required for this setup?

Francois-Esquire: Yes, we can use the features listed on both sources! I based the validation to account for all the keys present on https://developer.mozilla.org/en-US/docs/Web/Manifest , although key difference is we only require the `"name"` field and not necessarily `"icons"` at the moment. We can wrap the `icon` validation in `createIsRequired` if we desire.

prod-sample/sample-modules/frank-lloyd-root/0.0.3/src/pwa.js
Francois-Esquire: Good call, I'll remove them in favor of config/state driven urls.

docs/api/modules/App-Configuration.md
nellyk: Awesome, thank you! This looks amazing. Great work Tobia

nellyk: Maybe we can mention that `name` is required

src/server/middleware/pwa/validation.js
```diff
@@ -14,24 +14,149 @@
  * permissions and limitations under the License.
  */
```
mtomcal, line 15: This might be the start of a very lightweight validation library as alternative to other libraries Ive seen already implementing this similar logic like https://github.com/25th-floor/spected#advanced-example. This doesnt seem especially PWA specific this logic and could be used elsewhere as in for other appConfig validations going forward? How about moving this to something like `src/server/validate-shape/validations.js` perhaps?

Francois-Esquire: @mtomcal I can def see that this can be used by other parts of the app when it comes to checking/validation - could we give more thought to the idea first before moving it? Maybe only move over the helper functions with some preset shapes used - `const { validateWebManifest, isString, createIsArrayOf, createValidator } = require('../')`?

@JAdshead A `WebManifest` TS type would certainly make that runtime validation obsolete, I think the main perk of the config is it's meant to log what's expected from the manifest object shape when devving so the user can fix it on the go.

mtomcal: @JAdshead I think runtime validation is essential however, the generic nature of these seems like they need to be organized differently. We will come accross more situations where we want to provide detailed validations for any new App Configuration shapes and these will come in handy unless we remove these in favor of library that specializes in this kind of validation.

@Francois-Esquire Would you investigate into lightweight validation library alternatives we have used previously (like ajv) or some other library that would save us from maintaining our own validation tooling?

mtomcal: We decided as a team to punt the validation library conversation until later. Resolving this convo.

PR #123, opened 2020-04-28
feature: update prom-client

PixnBits:
## Description
`prom-client` 12.x added more metrics which cover those that we added via `prometheus-gc-stats`. Additionally, `prom-client` is making use of the new [`perf_hooks`](https://nodejs.org/docs/latest-v12.x/api/perf_hooks.html#perf_hooks_performanceentry_kind) API in Node.JS removing the need for `gc-stats` (binary dep). This, like #45 will help reduce potential install problems and binary incompatibilities (ex: switching base docker images).

## Motivation and Context
This keeps One App up-to-date and reduces installation complexity.

## How Has This Been Tested?
Integration tests have been added showing the metric names exposed.

## Types of Changes
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Probably a breaking change as the metric names change (and reducing names but increasing labels for some of that information). I'm hoping this can be added before v5.0.0 is officially released.

## Checklist:

This reduces one of the two native dependencies and so hopefully makes installation easier.


__tests__/integration/__snapshots__/one-app.spec.js.snap
```diff
@@ -42,6 +42,44 @@ Object {
 }
 `;
 
+exports[`Tests that require Docker setup one-app successfully started metrics has all metrics 1`] = `
```
PixnBits, line 45: I'd encourage looking at the diff of this file across the two commits to see the difference in metric names

__tests__/integration/__snapshots__/one-app.spec.js.snap
```diff
@@ -42,6 +42,44 @@ Object {
...
+  "nodejs_active_handles",
+  "nodejs_active_handles_total",
+  "nodejs_active_requests",
+  "nodejs_active_requests_total",
+  "nodejs_eventloop_lag_max_seconds",
+  "nodejs_eventloop_lag_mean_seconds",
+  "nodejs_eventloop_lag_min_seconds",
+  "nodejs_eventloop_lag_p50_seconds",
+  "nodejs_eventloop_lag_p90_seconds",
+  "nodejs_eventloop_lag_p99_seconds"
```
PixnBits, line 56: [from Travis](https://travis-ci.org/github/americanexpress/one-app/jobs/680785582):
```
    -   "nodejs_eventloop_lag_p99_seconds"
    +   "nodejs_eventloop_lag_p99_seconds",
```
🤦

PixnBits: as there aren't any reviews yet and we might want to keep the two commits separate I've replaced 39b35b5 with 3b23458 to reduce the need for squashing later
(might be moot if the metric test isn't desired on it's own, but just in case)

prod-sample/docker-compose.yml
```diff
@@ -14,6 +14,7 @@ services:
     image: one-app:at-test
     expose:
       - "8443"
+      - "3005"
```
JamesSingleton, line 17: I know it is the sample, but what expose 3005 now?

__tests__/integration/helpers/testRunner.js
```diff
@@ -46,7 +46,8 @@ const setUpTestRunner = async ({ oneAppLocalPortToUse } = {}) => {
           'one-app': {
             ports: [
               `${oneAppLocalPortToUse}:8443`,
-            ],
+              oneAppMetricsLocalPortToUse ? `${oneAppMetricsLocalPortToUse}:3005` : undefined,
```
infoxicator, line 49: @JamesSingleton the 3005 port has to be exposed because it is mapped here like from ramdomPort to 3005

__tests__/integration/one-app.spec.js
```diff
@@ -85,6 +88,31 @@ describe('Tests that require Docker setup', () => {
...
+      });
+
+      it('has help information on each metric', async () => {
+        expect.assertions(1);
+        const response = await fetch(appAtTestUrls.fetchMetricsUrl);
+        const parsedMetrics = parsePrometheusTextFormat(await response.text());
+        const errors = parsedMetrics
+          .filter((metric) => (!metric.help) || metric.help.length === 0)
+          .map((metric) => metric.name);
+        expect(errors).toEqual([]);
```
mtomcal, line 112: `it('has help information on each metric', async () => {` asserts `expect(errors).toEqual([]);`. It seems this test should be renamed to something related to `it('has no errors')` instead?

infoxicator: After entering @PixnBits brain, he tends to test the negative scenario and if it can't find anything then it is a successful test, the errors keyword is generic there I think 🤔

PixnBits: My thought was showing all, say, 20 errors at once would be easier to see what's going on, rather than stopping on the first bad assertion and not trying/reporting the other 19 issues.

mtomcal: I see your reasoning. Any chance we can update the assertion text to be more descriptive related to your point? If I were to maintain or update `parsePrometheusTextFormat` I wouldn't be sure what this test case's purpose.

PixnBits: as discussed I'll update this to test for the positive scenario 👍

PixnBits:
```
    Snapshot name: `Tests that require Docker setup one-app successfully started metrics has all metrics 1`

    - Snapshot
    + Received

    @@ -1,6 +1,8 @@
      Array [
    +   "circuit",
    +   "circuit_perf",
```
ah, #111 was merged! I ran the tests before rebasing but not after 🤦‍♂️

infoxicator:
> ```
>     Snapshot name: `Tests that require Docker setup one-app successfully started metrics has all metrics 1`
> 
>     - Snapshot
>     + Received
> 
>     @@ -1,6 +1,8 @@
>       Array [
>     +   "circuit",
>     +   "circuit_perf",
> ```
> 
> ah, #111 was merged! I ran the tests before rebasing but not after 🤦‍♂️

I thought I updated it but, might have messed up because I did it from my fork

PR #131, opened 2020-05-06
feat(pwa/caching): caching and invalidation

Francois-Esquire:
Extends the service worker with caching on the `fetch` event.

## Description

Using the `fetch` event, we have a caching workflow that invalidates and expires cache entries added to the cache. The caching targets `one-app` assets (including `modern`, `legacy` and `intl`), modules (code-split chunks too) and lang packs.

## Motivation and Context

This change will enable the service worker to respond to document requests for `.js, .json` resources commonly used in the ecosystem. As an enhancement, we become more network independent and moving towards offline-ready support.

## How Has This Been Tested?

Unit tests.

## Types of Changes

## Checklist:

One App users will be able to utilize the progressive enhancement if PWA is enabled and supported on client.


src/client/service-worker/index.js
```diff
@@ -15,13 +15,20 @@
  */
 
 export function importServiceWorkerClient(settings) {
-  // In the future, we should consider making one-service worker added dynamically
-  // as an external chunk with the ability to support conditional load.
   return import(/* webpackChunkName: "service-worker-client" */ './client')
+    // get the default export
     .then((imported) => imported.default)
+    // invoke with settings
     .then((serviceWorkerClient) => serviceWorkerClient(settings));
```
mtomcal, line 22: Why not remove the extra `then` handler?
```suggestion
return import(/* webpackChunkName: "service-worker-client" */ './client')
    // get the default export and invoke with settings
    .then(({ default: serviceWorkerClient }) => serviceWorkerClient(settings)
```

mtomcal, line 22: Why not remove the extra `then` handler?
```suggestion
return import(/* webpackChunkName: "service-worker-client" */ './client')
    // get the default export and invoke with settings
    .then(({ default: serviceWorkerClient }) => serviceWorkerClient(settings)
```

docs/api/modules/App-Configuration.md
```diff
@@ -234,6 +234,14 @@ if (!global.BROWSER) {
 }
 ```
 
+### Caching
+
+When the service worker is enabled, the Caches API is working on every `fetch` event
+that is intercepted by the service worker and caches/responds if they are either app,
+module or lang pack resources (the expected paths are the build outputs from `one-amex-bundler`).
```
nellyk, line 241:
```suggestion
module or lang pack resources (the expected paths are the build outputs from [`one-app-bundler`](https://github.com/americanexpress/one-app-cli/tree/master/packages/one-app-bundler)).
```

docs/api/modules/App-Configuration.md
```diff
@@ -188,7 +188,7 @@ the event we want to purge the service worker and remove any existing service
 workers on browser clients, We can use the escape hatch worker by setting
 `escapeHatch` to true.
 
-While in normal circumstances we would normally want to use the standard `serviceWorker`,
+While in normal circumstances we would want to use the standard `serviceWorker`,
 `recoveryMode` and `escapeHatch` are there to help the origin that `one-app` is deployed on;
```
mtomcal: Why a semicolon?

Francois-Esquire:
> What about integration tests for testing cache invalidation of modules?

I'm working on them, they will be included in this PR. Now that we got #135 in, should make lang packs also testable.

mtomcal:
Would you prefix this PR title with "WIP" then until you're ready to go with Integration tests?

Francois-Esquire:
> Would you prefix this PR title with "WIP" then until you're ready to go with Integration tests?

I've updated with integration tests!

prod-sample/sample-modules/frank-lloyd-root/0.0.3/package.json
```diff
@@ -20,12 +20,20 @@
     "react": "^16.4.2",
     "react-helmet": "^5.2.1",
     "react-immutable-proptypes": "^2.1.0",
+    "react-intl": "^3.6.0",
```
JamesSingleton, line 23: What is the reasoning for adding `react-intl` it isn't used within `frank-lloyd-root`

__tests__/client/service-worker/events/caching.spec.js
```diff
@@ -388,8 +388,10 @@ describe('caching', () => {
     }, Promise.resolve([]));
 
     expect(fetch).toHaveBeenCalledTimes(3);
-    events.forEach((event) => {
-      expect(event.waitUntil).toHaveBeenCalledTimes(3);
+    events.forEach((event, index) => {
+      // on the third call, we expect to waitUntil the original cache item was removed
+      // as an additional step when invalidating
+      expect(event.waitUntil).toHaveBeenCalledTimes(index === 2 ? 4 : 3);
```
mtomcal, line 394: This isnt clear to me. Can you explain this in more detail why the third call is important or what script the 3rd call is referring to?

Francois-Esquire: The reason for the difference in the third call is that `event.waitUntil` is called an extra time to wait for the invalidation to delete the item. I'll add comments, and maybe rethink the test to make it clearer.

__tests__/integration/one-app.spec.js
```diff
@@ -901,6 +902,236 @@ describe('Tests that require Docker setup', () => {
...
+                        // eslint-disable-next-line implicit-arrow-linebreak, no-confusing-arrow
+                        cache.keys().then((requests) => [
+                          key,
+                          requests.map(({ url }) => ({ url, cache: key })),
+                        ])
+                      )
+                  )
+                )
+              )
+              .then(done);
```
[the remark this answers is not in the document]
Francois-Esquire: You are absolutely right, it is really hard to read.. the primary reason I didn't use `async`/`await` syntax is babel will transform this within the test (unaware of different execution context) and being that it runs in the browser context via `executeAsync`, the transforms applied + helpers done by babel are not present and cause this function to fail.

I really would like to see this function simplified - looks like inline comments to disable babel from transforming certain parts of source files is not supported... I could break this up within the scope of the fn to simplify. Or possibly write it with `async`/`await` as a string... Any ideas?

prod-sample/sample-modules/frank-lloyd-root/0.0.3/package.json
Francois-Esquire: For the integration tests, I'm using `cultured-frankie` v`0.0.1` and does require `react-intl` as a `requiredExternal`, so I've added it here.

__tests__/integration/one-app.spec.js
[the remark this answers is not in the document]
Francois-Esquire: Per discussion, we agreed to track this for another effort.

docs/api/modules/App-Configuration.md
```diff
@@ -286,6 +286,14 @@ if (!global.BROWSER) {
 }
 ```
 
+### Caching
+
+When the service worker is enabled, the Caches API is working on every `fetch` event
```
mtomcal, line 291: So we can safely say that this caching functionality is completely "revertible" and opt-in along just like the core PWA feature is today?

__tests__/integration/one-app.spec.js
```diff
@@ -932,13 +933,269 @@ describe('Tests that require Docker setup', () => {
...
+            function getCacheKeys(done) { caches.keys().then(done); })
+          ).resolves.toEqual([]);
+        });
+      });
+
+      describe('caching', () => {
+        const oneAppVersionRegExp = /\/([^/]+)(?:\/i18n)?\/[^/]*\.js$/;
+
+        /* eslint-disable implicit-arrow-linebreak */
+        function _getCacheEntries(done) {
```
mtomcal, line 953: As discussed, add a `// TODO` here to refactor later.

docs/api/modules/App-Configuration.md
Francois-Esquire: Yes, when removing the service worker, the cache is completely cleared, restarting the service worker in recovery mode and back, it will trigger a reset of the cache. It isn't an opt-in is turned on automatically.

__tests__/integration/one-app.spec.js
Francois-Esquire: updated dd34146

docs/api/modules/App-Configuration.md
mtomcal: Gotcha 👍

prod-sample/sample-modules/frank-lloyd-root/0.0.3/src/pwa.js
```diff
@@ -27,17 +31,20 @@ export const webManifest = (config) => ({
   theme_color: '#FDB92D',
   icons: [
     {
-      src: `${config.cdnUrl}modules/frank-lloyd-root/0.0.3/assets/pwa-icon-180px.png`,
+      src: `${baseUrl}assets/pwa-icon-180px.png`,
```
JAdshead, line 34: whats the reason for not using the `config.cdnUrl` ?

prod-sample/sample-modules/frank-lloyd-root/0.0.3/src/pwa.js
```diff
@@ -27,17 +31,20 @@ export const webManifest = (config) => ({
...
+      // src: `${config.cdnUrl}modules/frank-lloyd-root/0.0.3/assets/pwa-icon-180px.png`,
```
JAdshead, line 35: can remove comments

src/client/service-worker/events/caching.js
```diff
@@ -0,0 +1,148 @@
...
+// -> we use /[^/]+/ to accurately capture what in between the slashes through negation ([^]).
+// -> for each named capture, we want to get specific on what we expect that match to be.
+// - we have a match-all in the center of the string (/https?(?::\/\/)HERE .* HERE\/(?<name>)/)
+// -> because it allows us to ignore any matches before the url is
+// -> matched from the end of the string and back
+// - if you run across `/(?<name> .{1,5})\1/`, the `\1` matches what was first matched
+// -> in a capture group. for this example: `name = \1`
+
+// groups: (modules/sha?) name / version / resource . bundle ? revision
+const moduleRegExp = /^https?(?::\/\/).*\/modules\/(?:[a-z0-9]+\/)?(?<name>[^/]+)\/(?<version>[^/]+)\/(?<resource>([a-zA-Z-~]+|\1)(?:\.\1)?)(?:\.chunk)?\.(?<bundle>(legacy\.)?browser)\.js(?:\?clientCacheRevision=(?<revision>[^/&]*))?$/;
```
JAdshead, line 44: I don't think we can rely on modules being hosted under a module directory. This would have to be documented as a requirement for running one-app and modules in production which i think defeats the strength of having urls within the module map.

The same comment goes for the other RegExps

src/client/service-worker/events/caching.js
```diff
@@ -0,0 +1,148 @@
...
+// groups: (modules/sha?) name / version / locale / resource
+const langPackRegExp = /^https?(?::\/\/).*\/modules\/(?:[a-z0-9]+\/)?(?<name>[^/]+)\/(?<version>[^/]+)(?:\/locale)?\/(?<locale>(?<language>[a-z]{2,3})(?:-)?(?<country>[a-zA-Z]{1,})?)\/(?<resource>(\1|[^/]*))\.json$/;
+// groups: (app) version / bundle / name
+const oneAppRegExp = /^https?(?::\/\/).*\/app\/(?<version>[^/]+)?\/(?:(?<bundle>legacy)\/)?(?:i18n\/)?(?<name>[^/]+)\.js$/;
```
JAdshead, line 48:
```suggestion
const oneAppRegExp = /^https?(?::\/\/).*\/app\/(?<version>[^\/]+)?\/(?:(?<bundle>legacy)\/)?(?:i18n\/)?(?<name>[^\/]+)\.js/;
```

PR #139, opened 2020-05-13
refactor(tests/integration): use alpine image

Francois-Esquire:
Update docker image to use `node:12-alpine` instead of `node:latest`.

## Motivation and Context

To add consistency with the other images we use (`v12`) and reduce downloads of 900+mb from the full image (with `alpine` distro).

## How Has This Been Tested?

Standard testing in place

## Types of Changes

## Checklist:


prod-sample/api/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:latest
+FROM node:12-alpine
```
infoxicator, line 1: Not related to this change, but we were going to check if one-app runs on 14 need to try it out

Francois-Esquire: That's a good idea!

PR #146, opened 2020-05-19
docs(appconfig): update app config docs

JamesSingleton:
## Description

Removes `appCompatibility` from the docs as it does not belong within appConfig, it should be in the bundler.

## Motivation and Context

So that end users are not confused on where to add app compatibility.

## How Has This Been Tested?

N/A

## Types of Changes

## Checklist:

### DIFF
--- a/docs/api/modules/App-Configuration.md
+++ b/docs/api/modules/App-Configuration.md
@@ -20,8 +20,6 @@ if (!global.BROWSER) {
     /* Child Module Specific */
     validateStateConfig,
     requiredSafeRequestRestrictedAttributes,
-    /* All Modules */
-    appCompatibility,
   };
 }
 ```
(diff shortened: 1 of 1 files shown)


JAdshead:
For reference, 

[Bundler adds the app compatibility to the appConfig](https://github.com/americanexpress/one-app-cli/blob/696609c702b128ba0339064173ac328ce8c00766/packages/one-app-bundler/webpack/loaders/validate-one-app-compatibility-loader.js#L18-L28)


Then [One App then uses app compatibility from app config to validate when module loaded](https://github.com/americanexpress/one-app/blob/master/src/server/utils/onModuleLoad.js#L86-L95), 

`appConfig.appCompatibility` was not intended to be provided by modules appConfig

PR #162, opened 2020-05-27
test(integration/pwa): inconsistent active state string

Francois-Esquire:
Removes a flaky integration test where the `{ active: { state: 'activated' } }` but often comes back as `registration.active.state === 'activating'`.

## Types of Changes

## Checklist:

None


10xLaCroixDrinker:
Since this is not a fix to the app, but to a test, the commit type should be `test` not `fix`

Francois-Esquire:
> Since this is not a fix to the app, but to a test, the commit type should be `test` not `fix`

I've amended the commit to use `test`

PR #172, opened 2020-05-28
chore(travis): run only two jobs

Francois-Esquire:
Only test unit & integration once per nodejs version (12.x, 14.x).

## Types of Changes

## Checklist:

Faster PR experience with `one-app`


JamesSingleton:
Just an FYI on why it runs the tests by default is because the default script is `npm test`.

PR #174, opened 2020-06-01
feat(make-promises-safe): treat unhandledRejection as uncaught exception

JAdshead:
## Description

Installed and imported [make-promises-safe](https://github.com/mcollina/make-promises-safe)

## Motivation and Context

Node does not crash if there is an 'unhandledRejection', i.e. a Promise without a .catch() handler. This causes a memory leak, make-promises-safe installs an process.on('unhandledRejection') handler that prints the stacktrace and exits the process with an exit code of 1, just like any uncaught exception.

## How Has This Been Tested?

Locally running one-app serving frank-lloyd-root. Unit and integration tests.

## Types of Changes

- [ ] Dependency update

## Checklist:

None

### DIFF
--- a/src/server/init.js
+++ b/src/server/init.js
@@ -14,6 +14,7 @@
  * permissions and limitations under the License.
  */
 
+import 'make-promises-safe';
 import 'core-js/stable';
 import 'regenerator-runtime/runtime';
 import 'cross-fetch/polyfill';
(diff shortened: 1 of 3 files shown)


src/server/init.js
```diff
@@ -17,6 +17,7 @@
 import 'core-js/stable';
 import 'regenerator-runtime/runtime';
 import 'cross-fetch/polyfill';
+import 'make-promises-safe';
```
infoxicator, line 20: Cool that works just by importing the module!

PR #181, opened 2020-06-10
fix(server-side-render): correct client mismatch

Francois-Esquire:
Fixes browser error message on rendering mismatch between server and client.
It should prevent any server side state config from leaking on the client (if config was used while rendering the html fragment middleware).

## Motivation and Context

Secure server only config from being used with rendering and remove the error on the browser console.

## How Has This Been Tested?

Locally validated and tests.

## Types of Changes

- [ ] Dependency update

## Checklist:

Removes browser error message on rendering mismatch and secures One App.


src/server/middleware/createRequestStore.js
```diff
@@ -58,6 +59,9 @@ export default function createRequestStore(
         extraThunkArguments: { fetchClient },
       });
 
+      // re-set the client config before rendering the html fragment/document
+      store.dispatch(setConfig(clientConfig));
```
JAdshead, line 63: this would mean that the client config is used for the server side render and negates the purpose of having a server config

JAdshead:
as mentioned in comment this negates the purpose of having a server config, therefore i am closing.

PR #194, opened 2020-06-11
fix(send-html): ignore async chunks

Francois-Esquire:
Does not include `service-worker-client` as a rendered script tag.

## Description

`service-worker-client` is intended to be an async script but is being rendered with SSR. Removing from the chunks will allow the script to be loaded client-side only.

## Motivation and Context

For more optimal initial loading, `service-worker-client` script was meant to be asynchronous, this fixes it from being loaded on initial load.

## How Has This Been Tested?

Local validation and testing suite

## Types of Changes

- [ ] Dependency update

## Checklist:


src/server/middleware/sendHtml.js
```diff
@@ -43,7 +43,7 @@ function getChunkAssets(assetsByChunkName) {
   return Object
     .entries(assetsByChunkName)
     // i18n is different per request, app needs to be the last chunk loaded
-    .filter(([chunkName]) => !chunkName.startsWith('i18n/') && chunkName !== 'app')
+    .filter(([chunkName]) => !chunkName.startsWith('i18n/') && !['app', 'service-worker-client'].includes(chunkName))
```
JAdshead, line 46: can you add a test to show this works as expected

Francois-Esquire: I have updated 1ed88f7

PR #199, opened 2020-06-15
feat(createTimeoutFetch): replaced with @americanexpress/fetch-enhancers

JAdshead:
## Description

Replace one-app's `createTimeoutFetch` provided to the client with [@americanexpress/fetch-enhancers/createTimeoutFetch](https://github.com/americanexpress/fetch-enhancers#createtimeoutfetch-server--browser)
## Motivation and Context

`@americanexpress/fetch-enhancers/createTimeoutFetch` takes advantage of the AbortController and will terminate timedout requests rather than by throwing an error.

## How Has This Been Tested?

Unit and integration tests.

## Types of Changes

- [ ] Dependency update

## Checklist:


package.json
```diff
@@ -74,6 +74,7 @@
   ],
   "dependencies": {
     "@americanexpress/env-config-utils": "^2.0.2",
+    "@americanexpress/fetch-enhancers": "^1.0.0-rc1",
```
mtomcal, line 77: If we're using this with One App, shouldn't we be out of `rc` with `fetch-enhancers`?

JAdshead: yeah thats a fair point, i don't see anything changing but can hold off until `1.0.0`

JAdshead: updated

src/universal/utils/transit.js
```diff
@@ -20,7 +20,6 @@
 
 import transit from 'transit-immutable-js';
 import { serializeError } from 'serialize-error';
-import { TimeoutError } from './createTimeoutFetch';
```
JamesSingleton, line 23: So `transit.js` no longer needs a TimeoutError? Is that because it is handled in prerender.js

JAdshead: So TimeoutError is still an instance of Error so should still be handled, createTimeoutFetch also uses abort controller rather and falling back to the timeout error if abort controller fails, this should never happen especially on the server.

JAdshead: So TimeoutError is still an instance of Error so should still be handled, createTimeoutFetch also uses abort controller rather and falling back to the timeout error if abort controller fails, this should never happen especially on the server.

PR #200, opened 2020-06-16
feat(pwa/offline): html shell

Francois-Esquire:
Adds offline navigation support for One App instances when service worker is enabled.

## Description

This PR includes an additional middleware `src/server/middleware/pwa/offline.jsx` and adds the `fetch` event to the service worker with `src/client/service-worker/events/fetch.js`.
To support offline navigation, `__render_mode__` was added to instruct the client side how to render the `react` root with either `hydrate` or `render` as the two options.

## Motivation and Context

The changes in this PR allow users to navigate while offline or when the network is unavailable. When coupled with future caching support, One App will be fully offline-ready out of the box.

## How Has This Been Tested?

It has been validated locally and through testing suite.

## Types of Changes

- [ ] Dependency update

## Checklist:

Offline capabilities when navigating with One App.


src/server/middleware/pwa/offline.jsx
```diff
@@ -0,0 +1,70 @@
...
+import createRequestStore from '../createRequestStore';
+import sendHtml from '../sendHtml';
+
+import { getServerPWAConfig } from './config';
+
+async function appShellMiddleware(req, res, next) {
+  const initialState = req.store.getState();
+  const rootModuleName = initialState.getIn(['config', 'rootModuleName']);
+
+  await req.store.dispatch(composeModules([{ name: rootModuleName }]));
```
mtomcal, line 33: Simplify to `const { store: { getState, dispatch } } = req`?

src/server/middleware/pwa/offline.jsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+function composeMiddleware(middlewareStack) {
+  return async function composition(req, res, next) {
+    await (async function stackIterator(stack) {
+      const middleware = stack.shift();
+      await middleware(req, res, stack.length === 0 ? next : stackIterator.bind(null, stack));
+    }([...middlewareStack]));
+  };
+}
```
mtomcal, line 55: Can we find a simpler method to loading async sources before wrapping a middleware function? This seems unnecessarily complex and will cause issue for those maintaining this. What about breaking up the compose and the async loads into two separate steps to make this simpler and more maintainable. For some this will be a black box, let's avoid this.

src/server/middleware/pwa/offline.jsx
```diff
@@ -0,0 +1,70 @@
...
+
+export default function createOfflineMiddleware(oneApp) {
+  const middlewareStack = composeMiddleware([
+    addFrameOptionsHeader,
+    createRequestStore(oneApp),
+    appShellMiddleware,
+    sendHtml,
+  ]);
```
mtomcal, line 63: Out of all these middlewares, how many of these require async?

src/server/middleware/pwa/offline.jsx
mtomcal: Also, where is this specific function tested? Given the complexity, there needs to be standalone tests for this piece.

src/server/middleware/pwa/offline.jsx
Francois-Esquire: Only one requires async `appShellMiddleware`

src/server/middleware/pwa/offline.jsx
Francois-Esquire: There are no async loading really and the `async/await` is only for `composeModules` usage in `appShellMiddleware`. I've simplified to only the recursion of the middleware layers in the stack - the function itself goes one middleware at a time and uses the `next` call to trigger the next middleware.

src/server/middleware/pwa/offline.jsx
```diff
@@ -0,0 +1,70 @@
...
+  return async function composition(req, res, next) {
+    await (async function stackIterator(stack) {
+      const middleware = stack.shift();
+      await middleware(req, res, stack.length === 0 ? next : stackIterator.bind(null, stack));
+    }([...middlewareStack]));
+  };
+}
+
+export default function createOfflineMiddleware(oneApp) {
+  const middlewareStack = composeMiddleware([
```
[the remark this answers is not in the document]
Francois-Esquire: I have simplified and peeled a few layers off d992985

src/server/middleware/pwa/offline.jsx
```diff
@@ -0,0 +1,70 @@
...
+    addFrameOptionsHeader,
+    createRequestStore(oneApp),
+    appShellMiddleware,
+    sendHtml,
+  ]);
+
+  return async function offlineMiddleware(req, res, next) {
+    const { serviceWorker } = getServerPWAConfig();
+    if (serviceWorker === false) next();
```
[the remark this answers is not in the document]
Francois-Esquire: That could lead to bugs I believe - for instance, if someone enabled service worker and deployed the app they would start with this middleware enabled. If they decide to disable the service worker, it would remain enabled until the app is redeployed

src/client/service-worker/events/fetch.js
```diff
@@ -0,0 +1,39 @@
...
+  appShell,
+  cacheRouter,
+  expiration,
+  createMiddleware,
+  createCacheName,
+} from '@americanexpress/one-service-worker';
+
+import { OFFLINE_CACHE_NAME } from '../constants';
+
+export default function fetchMiddleware() {
```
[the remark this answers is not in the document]
Francois-Esquire: I have updated d992985

src/server/middleware/pwa/offline.jsx
Francois-Esquire: I have updated d992985

docs/api/modules/App-Configuration.md
```diff
@@ -233,11 +233,22 @@ For the variety of service workers available, we have control to set its
...
 be a `Function` and is passed the `clientConfig` as the only argument.
 
+#### App Install and Offline Capability
+
+One App supports offline navigation capabilities when the network is unavailable and the
+service worker is enabled. With a configured web app manifest and root module, One App
+clients can be installed as a PWA across devices and platforms.
+To enable installing an app, please set the value for `start_url` in the web manifest
+to load the route that is desired when an installed PWA is opened directly from the device;
+for the best experience, consider making a Holocron module route to match the `start_url`.
```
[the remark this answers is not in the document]
Francois-Esquire: Good call, I have reworded fb2ef19

src/server/middleware/pwa/offline.jsx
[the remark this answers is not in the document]
mtomcal: Do you mean doing an early return if `serviceWorker` is false, @JAdshead?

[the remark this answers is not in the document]
JAdshead: now there are not so many layers it does not make sense, i guess i missed what the composeMiddleware was retuning

src/client/initClient.jsx
```diff
@@ -63,7 +63,9 @@ export default async function initClient() {
     );
     /* eslint-enable react/jsx-props-no-spreading */
 
-    hydrate(
+    const { __render_mode__: renderMode = 'hydrate' } = global;
```
JAdshead, line 66: this should probably be a bit more defensive, if `__render_mode__` is somehow set by something else could cause issues
```suggestion
    const renderMode = global.__render_mode__ === 'render'  ?  'render' : 'hydrate'
```

Francois-Esquire: I have updated aad162a

PR #206, opened 2020-06-19
fix(module-map): missing baseUrl to module map on server

infoxicator:
Fixes an issue where the `baseUrl` key was present on the moduleMap cached sent to the client but not present on the server module map.

## Description

 The baseUrl key is used by `one-app-ducks` to determine the url where the language pack statics are stored. However, when doing Server Side Rendering, the baseUrl is not present and it was falling back to the client rendering to load the language pack (breaking server side rendering) when using either `loadLanguagePack` or `queryLanguagePack`

## Motivation and Context

This change fixes this issue by taking the `node` bundle type for each module and creating the `baseUrl` from the `url` in the `node` bundle key.

## How Has This Been Tested?

This is a non-breaking bugfix that solves an issue with `loadLanguagePack` 

This solution uses the browser key in the module map which is available on both the clientModuleMapCache and the moduleMap already available on the server.

Tested this solution with `queryLanguageData` and a clean module generated with `intl`, the language pack now loads on the server on the first render

Component rerender: 1
LangpackTest.jsx:16 languageData =>  {locale: "en-US", greeting: "Welcome to langpack-test from the United States!"}
LangpackTest.jsx:17 isLoading =>  false

Tested with cultured-frankie and `loadLanguageData` however, even though the language pack loads on the server on the first render, the `moduleStatus` property only changes from loading to loaded on the second render, this has been identified as a separate issue and will be addressed on a different PR.

## Types of Changes

- [ ] Dependency update

## Checklist:

No expected impact


src/server/utils/addBaseUrlToModuleMap.js
```diff
@@ -0,0 +1,32 @@
...
+/* The baseUrl key in the module map is used by one-app-ducks to construct
+the language pack URL pointing at the location of the language pack */
+const addBaseUrlToModuleMap = (moduleMap) => ({
+  ...moduleMap,
+  modules: Object.entries(moduleMap.modules).reduce((acc, [moduleName, moduleBundles]) => (
+    {
+      ...acc,
+      [moduleName]: {
+        ...moduleBundles,
+        baseUrl: moduleBundles.baseUrl ? moduleBundles.baseUrl : moduleBundles.node.url.replace(/[^/]+\.js$/i, ''),
```
10xLaCroixDrinker, line 26: this option should be documented

PR #212, opened 2020-06-24
feat(pwa/cache): clear when disabled

Francois-Esquire:
Clears the Cache Storage when the service worker is disabled or in recovery mode.

## Description

The changes proposed happen mainly on the client in `src/client/service-worker/index.js`. They are run when the service worker is not enabled.

## Motivation and Context

When the service worker is activated, it currently initializes the cache with the `manifest.webmanifest` and the html shell. As we add module caching, we will have other active caches populated. The changes proposed clear the cache when the service worker is not active or in recovery mode.

## How Has This Been Tested?

Testing suite.

## Types of Changes

- [ ] Dependency update

## Checklist:

Removes any cache entries that pertain to One App.


__tests__/integration/one-app.spec.js
```diff
@@ -978,6 +985,13 @@ describe('Tests that require Docker setup', () => {
           });
 
           expect(result).toBe(null);
+
```
infoxicator, line 988: is it worth to add a tests that deletes only the keys cached by the service worker and nothing else, as I understand it, it should keep keys in the cache that don't have anything to do with our SW

__tests__/client/service-worker/index.spec.js
```diff
@@ -84,6 +90,22 @@ describe('initializeServiceWorker', () => {
...
+
+    expect(caches.snapshot()).toEqual({
+      'one-app-test': {
+        'http://localhost/index.js': null,
+      },
+    });
+    await expect(initializeServiceWorker({ serviceWorker: false })).resolves.toBe(registration);
+    expect(registration.unregister).toHaveBeenCalledTimes(1);
+    expect(serviceWorkerClient).not.toHaveBeenCalled();
+    expect(caches.snapshot()).toEqual({});
```
infoxicator, line 106: is it worth to add a tests that deletes only the keys cached by the service worker and nothing else, as I understand it, it should keep keys in the cache that don't have anything to do with our SW

Francois-Esquire: That's a good point. I'll update to add this.

Francois-Esquire: I have updated to filter out cache keys created by the service worker with the prefix `__sw` ( 2472120 )

PR #225, opened 2020-07-02
feat(pwa/cache): add resource caching and invalidation

Francois-Esquire:
Service worker caching and cache invalidation for app, module and lang-pack resources when the service worker is enabled.

## Description

Extends the capability of the `fetch` event to respond to One App and module resources from a `holocron` module map. It uses the `baseUrl` to cache module and lang-pack assets and the current One App version. Invalidation happens when either the `clientCacheRevision`, a resource version (both app and modules) or bundle type (modern or legacy) has changed. Locale change also triggers invalidation for module lang-packs.

We also tap into the `expiration` middleware (already present) to set the resources to be cached for only 30 days.

## Motivation and Context

Supports offline for all the essential resources to start One App client-side and provides network independence (predictable, fast response times with any network condition) for One App resources, modules and language packs.

## How Has This Been Tested?

Unit tests and local validation.

## Types of Changes

- [ ] Dependency update

## Checklist:

Offline capability and network predictability & speed.


__tests__/client/service-worker/worker.spec.js
```diff
@@ -40,6 +40,11 @@ beforeEach(() => {
   jest.clearAllMocks();
 });
 
+beforeAll(() => {
+  process.env.ONE_APP_BUILD_VERSION = '5.0.0';
```
JamesSingleton, line 44: Out of curiosity, should we just read from the `package.json` here instead of hardcoding it to `5.0.0`?

Francois-Esquire: For testing, I think that would work - do you foresee hardcoding causing any issues?

src/client/service-worker/events/utility.js
```diff
@@ -0,0 +1,109 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import {
+  match, put, remove, getMetaData, setMetaData, createCacheName,
+} from '@americanexpress/one-service-worker';
+
+const localeRegExp = /(?<locale>([a-z]{2,3})(?:-)?([a-zA-Z]{1,})?)\/[^/]*\.json/;
```
JAdshead, line 21: FYI this might cause us issues with the next eslint update which check regexs using - https://github.com/substack/safe-regex
maybe only having the one group will help simplify

src/client/service-worker/events/utility.js
```diff
@@ -0,0 +1,109 @@
...
+
+export function createResourceMetaData(event, resourceInfo, revision) {
+  const [name, baseUrl] = resourceInfo;
+  const [version] = baseUrl.replace(/\/$/, '').split('/').reverse();
+  const { request } = event;
+
+  let type = name === 'app' ? 'one-app' : 'modules';
+
+  let locale = null;
```
JAdshead, line 30: why just not leave it as ? 
```suggestion
  let locale;
```

src/client/service-worker/events/utility.js
```diff
@@ -0,0 +1,109 @@
...
+  if (localeRegExp.test(request.url)) {
+    type = 'lang-packs';
+    locale = request.url.match(localeRegExp).groups.locale;
+  }
+
+  let bundle = 'browser';
+  if (request.url.includes('legacy')) {
```
JAdshead, line 37: what if the module is called `legacy` so the bundle will be `legacy.browser.js`

infoxicator: far fetched scenario but def a possibility

Francois-Esquire: I've updated to add more specificity (fb12731)

src/client/service-worker/events/utility.js
Francois-Esquire: Good point, I've simplified the regexp (fb12731)

src/client/service-worker/events/utility.js
```diff
@@ -0,0 +1,111 @@
...
+
+  let path;
+  let locale;
+  if (localeRegExp.test(request.url)) {
+    type = 'lang-packs';
+    [path, locale] = request.url.match(localeRegExp);
+  }
+
+  let bundle = 'browser';
+  if (legacyRegExp.test(request.url)) {
```
JAdshead, line 39: can you add some tests around this ? looks like it will still fail for the following
* `cdn.com/modules/legacy/demo.browser`
* `cdn.com/modules/legacy.browser` even though there is a `cdn.com/modules/legacy.legacy.browser`

src/client/service-worker/events/utility.js
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+import {
+  match, put, remove, getMetaData, setMetaData, createCacheName,
+} from '@americanexpress/one-service-worker';
+
+const localeRegExp = /(([a-z]{2,3})(?:-)?([a-zA-Z]{1,})?)\/[^/]*\.json/;
+const legacyRegExp = /legacy\.browser|\/legacy\//;
+
+export function createResourceMetaData(event, resourceInfo, revision) {
```
JAdshead, line 24: I don't see any unit tests for these functions, this is really important especially when introducing complex regexs

src/client/service-worker/events/utility.js
```diff
@@ -0,0 +1,111 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import {
+  match, put, remove, getMetaData, setMetaData, createCacheName,
+} from '@americanexpress/one-service-worker';
+
+const localeRegExp = /(([a-z]{2,3})(?:-)?([a-zA-Z]{1,})?)\/[^/]*\.json/;
```
JAdshead, line 21: https://www.npmjs.com/package/safe-regex shows as unsafe, please test thoroughly.

src/client/service-worker/events/utility.js
```diff
@@ -0,0 +1,111 @@
...
+  if (revision) metaData.revision = revision;
+  if (locale) metaData.locale = locale;
+
+  return metaData;
+}
+
+export function markResourceForRemoval(cachedMetaRecord, newMetaRecord) {
+  if (cachedMetaRecord.revision !== newMetaRecord.revision) return true;
+  if (cachedMetaRecord.version !== newMetaRecord.version) return true;
+  if (cachedMetaRecord.bundle !== newMetaRecord.bundle) return true;
```
JAdshead, line 62: is this needed ? the url will differ between a legacy and modern module so they should never have a cache collision

__tests__/client/service-worker/worker.spec.js
JamesSingleton: Nope, I was just curious

src/client/service-worker/events/utility.js
Francois-Esquire: It isn't needed per se, however it does invalidate if the "same" version comes in, or a browser upgrades for example. It would then remove the previous bundle variant.

Francois-Esquire: If we don't care to do this, I can remove

src/client/service-worker/events/utility.js
Francois-Esquire: For `cdn.com/modules/legacy/demo.browser`, we need to carry a certain convention for a baseline and we can rely on the bundler extracts for the module file structure (eg `cdn.com/modules/demo/build/1.0.0/demo.browser.js`). If a folder is placed under `/legacy/` we can consider it so.

For `cdn.com/modules/legacy.browser`, I'll update the regexp to `/\.legacy\.browser\./` to expect `*.legacy.browser.*`.

src/client/service-worker/events/utility.js
Francois-Esquire: When it comes to safety concerns, let's establish that this is a client-side app and any attacks on the RegExp would only affect the attacker at this point. The RegExp has three capture groups and no forward/backward looking behavior - I will tighten it up and update with that.

The library does state "WARNING: This module has both false positives and false negatives. Use `vuln-regex-detector` for improved accuracy." which may not be accurate.

src/client/service-worker/events/utility.js
Francois-Esquire: The goal was to test the implementation, not the utility functions per se but that the system works and is covered. What do you think makes the RegExps complex?

Francois-Esquire: I've broken up the `regexp` and simplified it to one main capture group if necessary for matching (efd5586)

src/client/service-worker/events/utility.js
Francois-Esquire: Simplified the regexps in efd5586

src/client/service-worker/events/utility.js
Francois-Esquire: The other thing is we need to have some BCP 47 shaped regexp which is why it may be more involved.

JAdshead: I am not able to look at that regex and instantly understand all the patterns which would fail and match, you might be a regex wizard but not everyone who works on or will work on will be. These functions are perfect examples of where unit tests are necessary. The unit tests not only protect from regression but they help convey and show the intended api for the function. You need to test the implementation as a whole and at the level of each function.

src/client/service-worker/events/utility.js
JAdshead: we still needs tests to show these scenarios and what is expected.

src/client/service-worker/events/utility.js
JAdshead: from my understanding `bundle` is either `legacy` or` browser`, if it does not change then this will return false so it will not invalidate. also not sure how a browser upgrading will alter the result of this condition ?

src/client/service-worker/events/fetch.js
```diff
@@ -23,8 +23,50 @@ import {
...
+      context.set('cacheName', meta.cacheName);
+      context.set('request', event.request.clone());
+      event.respondWith(fetchCacheResource(event, meta));
+    }
+  };
+}
 
 export default function createFetchMiddleware() {
+  const oneAppVersion = process.env.ONE_APP_BUILD_VERSION;
+  const holocronModuleMap = JSON.parse(process.env.HOLOCRON_MODULE_MAP);
```
mtomcal, line 68: `process.env.HOLOCRON_MODULE_MAP` Where does this come from? Having trouble finding this. Where is the documentation on these ENV vars? https://one-amex-docs.americanexpress.com/en-us/one-app/api/server/environment-variables

Francois-Esquire: Both `ONE_APP_BUILD_VERSION` and `HOLOCRON_MODULE_MAP` are injected values; `ONE_APP_BUILD_VERSION` is [replaced during build time ](https://github.com/americanexpress/one-app/blob/feature/pwa-cache/scripts/build-service-workers.js#L31) and the `HOLOCRON_MODULE_MAP` is injected in the [service worker middleware](https://github.com/americanexpress/one-app/blob/feature/pwa-cache/src/server/middleware/pwa/service-worker.js#L21-L24).

Francois-Esquire:
> utility functions need unit testing.

@JAdshead unit tests have been added

src/client/service-worker/events/fetch.js
mtomcal: Could we place this information in a well thought out README? This process is rather unusual given the nature of merging a PWA-only build system into Server middleware.

__tests__/client/service-worker/events/utility.spec.js
```diff
@@ -0,0 +1,241 @@
...
+    ['https://example.com/cdn/modules/test-root/2.2.2/test-root.browser.js', moduleInfo, {
+      ...baseMetaData,
+      path: 'test-root.browser.js',
+      revision: null,
+    }],
+  ])('extracts metadata from %s', (url, [type, baseUrl], result) => {
+    // eslint-disable-next-line no-param-reassign
+    if (result.revision === null) delete result.revision;
+    const meta = createResourceMetaData({ request: { url } }, [type, baseUrl], result.revision);
+    expect(meta).toEqual({ ...result, url });
```
JAdshead, line 141: Sorry, this seems overly abstracted for a test, I am struggling to figure out given X the result should be Y. You can make tests as verbose as they need to be so it clearly shows the api is doing whats expected.

__tests__/client/service-worker/events/utility.spec.js
```diff
@@ -0,0 +1,241 @@
...
+    const event = { request: { url: meta.url }, waitUntil };
+    const response = {};
+    const responseHandler = invalidateCacheResource(event, meta);
+
+    getMetaData.mockImplementationOnce(() => Promise.resolve({}));
+
+    expect(responseHandler(response)).toBe(response);
+    expect(waitUntil).toHaveBeenCalledTimes(1);
+    expect(getMetaData).toHaveBeenCalledTimes(1);
+    expect(getMetaData).toHaveBeenCalledWith({ cacheName: 'modules/assets/cat.jpg', url: 'http://localhost/modules/assets/cat.jpg' });
```
JAdshead, line 162: `modules/assets/cat.jpg` looks like a path but it does not appear to be a valid path. thoughts on changing pattern to `modules-assets-cat.jpg` to make it clear that its the cache name not a path ? might be overkill.

JAdshead, line 162: `http://localhost/modules/assets/cat.jpg` does not appear to be a valid url. should this be the case ?

__tests__/client/service-worker/events/utility.spec.js
```diff
@@ -0,0 +1,241 @@
...
+    expect(put).toHaveBeenCalledTimes(1);
+    expect(put).toHaveBeenCalledWith('clone', 'clone', meta);
+  });
+});
+
+describe(fetchCacheResource.name, () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
```
JAdshead, line 209: these tests lean towards how the function is built more than what it's doing. 

`given match returns a response i expect to see the cached response and not call fetch.`

`given match does not return a cached response i expect fetch to have been called with X, set the result in the cache and ...`

src/client/service-worker/events/fetch.js
Francois-Esquire: Updated with additional docs breaking down details (f21a72c) (a4525b7)

__tests__/client/service-worker/events/utility.spec.js
Francois-Esquire: Valid point, I'll update this.

__tests__/client/service-worker/events/utility.spec.js
Francois-Esquire: So I mentioned [in this reply](https://github.com/americanexpress/one-app/pull/225#discussion_r463264375) about `cacheName` being a cache entry ID for the JSON resource meta data that is stored using the Cache API. It does need to be a path for it to be `match`ed.

__tests__/client/service-worker/events/utility.spec.js
Francois-Esquire: I'm not fully getting this point.. how would you test it?

Francois-Esquire: I made the `fetch.spec.js` to unit test the system/implementation as a whole, while `utility.spec.js` focuses on the portions of the system. Also, integration tests will cover the implementation.

__tests__/client/service-worker/events/utility.spec.js
Francois-Esquire: Updated (c48473f)

src/client/service-worker/events/utility.js
Francois-Esquire: Yes exactly, invalidation would remain false - it would only invalidate when the `browserlist` used to generate `last 2 versions` for modern bundles is updated or a user lags behind a few version updates and decides to update their browser, or if a user has to downgrade for whatever reason/restriction.

PR #257, opened 2020-07-27
chore(script): add cross-env to build:server

JamesSingleton:
## Description

This add `cross-env` to the `build:server` script so that it does not fail on a Windows machine

## Motivation and Context

So that building One App does not fail on a Windows machine

## How Has This Been Tested?

Tested on a Windows machine where the build was failing due to this.

## Types of Changes

- [ ] Dependency update

## Checklist:

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,7 @@
     "clean:test": "rimraf test-results",
     "prebuild": "npm run clean:build",
     "build:bundle": "cross-env NODE_OPTIONS=--max-old-space-size=4096 bundle-one-app",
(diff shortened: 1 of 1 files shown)


mtomcal:
Can you update the package-lock?

JamesSingleton:
> Can you update the package-lock?

@mtomcal no need. `cross-env` was already a `devDep`, I was just adding it to an existing script. The DangerJS script just checks for any change, not if it is a `dependency` or not.

PR #278, opened 2020-08-18
feat(release): move to github actions from travis

nellyk:
## Description

These introduces github actions for unit tests and integration tests and removes travis settings

## Motivation and Context

This would enable us leverage github actions for deployments and have all cicd working on github actions and leverage the [benefits](https://github.com/features/actions) that come with it. 

## How Has This Been Tested?

Please check the checks tab above to see the actions working
## Types of Changes

- [ ] Dependency update

## Checklist:

No impact


Dockerfile
```diff
@@ -39,7 +39,6 @@ EXPOSE 3002
 EXPOSE 3005
 WORKDIR /opt/one-app
 RUN chown node:node /opt/one-app
-USER node
```
infoxicator, line 42: 🤔 I thought we found a workaround to use node as the user instead of root, we need to investigate this change and ensure it doesn't cause any permission issues

nellyk: https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user due to this that wasn't possible. I did try a few different options 

> Docker actions must be run by the default Docker user (root). Do not use the USER instruction in your Dockerfile, because you won't be able to access the GITHUB_WORKSPACE

I tested out the docker images on a module locally and it worked fine. We might need context on why this was added.

.github/workflows/pull_request_integration_test .yml
```diff
@@ -0,0 +1,23 @@
+name: One App Integration Tests - PR
```
mtomcal, line 1: Theres a space character in `.github/workflows/pull_request_integration_test .yml`

Dockerfile
```diff
@@ -38,18 +40,22 @@ EXPOSE 3001
...
+USER $USER
 RUN chown node:node /opt/one-app
 CMD ["node", "lib/server"]
 COPY --from=builder --chown=node:node /opt/one-app/development ./
 
 # production image
 # last so that it's the default image artifact
 FROM node:12-alpine as production
+ARG USER
+ENV USER ${USER:-node}
```
mtomcal, line 52: `-node` is this the default fallback arg?

mtomcal:
This looks good implementing what we discussed last week 👍

mtomcal (dismissed):

Francois-Esquire (dismissed):

Dockerfile
nellyk: yes it is, if no value is provided it would default to it

infoxicator (approved):

PR #281, opened 2020-08-19
feat(sendHtml): allow custom error page

JamesSingleton:
## Description

This allows ONE_FALLBACK_URL to be given in order to define a custom
error page.

So this will require documentation on it, but I wanted to wait for approvals just in the off chance that the implementation changes due to reviews.

## Motivation and Context

Allow someone to override the default error page.

## How Has This Been Tested?

This was tested locally returning `renderStaticErrorPage(res)` and testing with different scenarios. Also, unit tests.

## Types of Changes

- [ ] Dependency update

## Checklist:

They can now provide `ONE_FALLBACK_URL` to be used as the error page instead of the default.


Francois-Esquire:
@JamesSingleton Could we add documentation for `ONE_FALLBACK_URL` and how we can use this feature?

JamesSingleton:
> @JamesSingleton Could we add documentation for `ONE_FALLBACK_URL` and how we can use this feature?

Yeah, I was going to wait to add the docs until after the reviews in the off chance that implementation changes based on feedback. I mentioned that in the description, I wish we could add "tasks" to be completed.

src/server/middleware/sendHtml.js
```diff
@@ -57,7 +57,23 @@ export function safeSend(res, ...payload) {
...
+      if (!contentType.includes('text/html')) {
+        return Promise.reject(new Error('Content-Type was not of type text/html'));
+      }
+      // Read the response as text.
+      return response.text();
+    })
+    .then((data) => safeSend(res, data));
+
+  return customErrorResponse;
+}
```
mtomcal, line 74:
```suggestion
export async function fetchCustomErrorPage(fallbackUrl, res) {
  const response = await fetch(fallbackUrl);
  // If the Content-Type is not text/html throw an error
  const contentType = response.headers.get('content-type');
  if (!contentType.includes('text/html')) {
    throw new Error('Content-Type was not of type text/html'));
  }
  // Read the response as text.
  const data = await response.text();
  return safeSend(res, data);
}
```

mtomcal: No need to mix Promise chaining with async await. I suggested the equivalent functionality to the promise chaining in pure async await. Hope this helps.

src/server/middleware/sendHtml.js
```diff
@@ -69,30 +85,36 @@ export function renderStaticErrorPage(res) {
...
-          <body style="background-color: #F0F0F0">
-            <div id="root">
-              <div>
-                <div style="width: 70%; background-color: white; margin: 4% auto;">
-                  <h2 style="display: flex; justify-content: center; padding: 40px 15px 0px;">Loading Error</h2>
-                  <p style="display: flex; justify-content: center; padding: 10px 15px 40px;">
-                    ${message}
-                  </p>
-                </div>
+  const fallbackUrl = process.env.ONE_FALLBACK_URL;
```
mtomcal, line 88: IIRC, we will need to have this environment variable validated alongside the other env variables.

mtomcal: https://github.com/americanexpress/one-app/blob/master/src/server/config/env/runTime.js

src/server/middleware/sendHtml.js
```diff
@@ -57,7 +57,19 @@ export function safeSend(res, ...payload) {
   }
 }
 
-export function renderStaticErrorPage(res) {
+export async function fetchCustomErrorPage(fallbackUrl, res) {
+  const response = await fetch(fallbackUrl);
```
mtomcal, line 61: This `fetch` has 6 sec timeout right?

JamesSingleton: That is a good question, it is the poly filled fetch from `cross-fetch`. Where would I see if there is a timeout?

mtomcal: I would add a 3 sec timeout https://github.com/americanexpress/fetch-enhancers#createtimeoutfetch-server--browser for quality.

src/server/middleware/sendHtml.js
```diff
@@ -57,7 +57,19 @@ export function safeSend(res, ...payload) {
...
+  // If the Content-Type is not text/html throw an error
+  const contentType = response.headers.get('content-type');
+  if (!contentType.includes('text/html')) {
+    throw new Error('Content-Type was not of type text/html');
```
infoxicator, line 65: Is it possible to fallback to the standard error page if something goes wrong fetching / parsing the custom url?

__tests__/server/middleware/sendHtml.spec.js
```diff
@@ -825,6 +828,62 @@ describe('sendHtml', () => {
...
+    it('returns default error page if response headers are not text/html', async () => {
+      process.env.ONE_FALLBACK_URL = 'https://example.com';
+      global.fetch = jest.fn(() => Promise.resolve({
+        headers: new global.Headers({
+          'Content-Type': 'text/plains',
+        }),
+      }));
+
+      await expect(renderStaticErrorPage(res)).rejects.toEqual(
+        new Error('Content-Type was not of type text/html')
```
infoxicator, line 884: Maybe I am missing something but after this error is thrown does it respond with the default error page? if it does we probably need to add that additional assertion here

infoxicator (changes requested):

src/server/config/env/runTime.js
```diff
@@ -199,6 +199,11 @@ const runTime = [
     normalize: (value) => value === 'true',
     defaultValue: () => false,
   },
+  {
+    name: 'ONE_FALLBACK_URL',
```
nellyk, line 203: could we please add some docs for this and how it can be used?

src/server/middleware/sendHtml.js
```diff
@@ -69,30 +81,36 @@ export function renderStaticErrorPage(res) {
     message = 'Sorry, we are unable to load this page at this time.';
```
nellyk, line 81: is it possible to have this custom too? especially with internationalization?

__tests__/server/middleware/sendHtml.spec.js
JamesSingleton: No it does not respond with the default

src/server/middleware/sendHtml.js
JamesSingleton: @mtomcal said that we should throw here

src/server/config/env/runTime.js
JamesSingleton: @nellyk as mentioned in the description of the PR, I was waiting for the approvals so that I knew the implementation was 100% before writing the docs. I didn't want to write the docs and then have to re-write them because something changed due to reviews.

src/server/middleware/sendHtml.js
JamesSingleton: Possibly in another PR, this PR is just to allow a custom error page.

__tests__/server/middleware/sendHtml.spec.js
infoxicator: I think we should fallback to the default error page, otherwise that error could be mistaken for a bad request and not that the error page was mis-configured.

__tests__/server/middleware/sendHtml.spec.js
```diff
@@ -97,7 +97,7 @@ jest.mock('../../../src/universal/utils/transit', () => ({
 }));
 
 jest.spyOn(console, 'info').mockImplementation(() => {});
-jest.spyOn(console, 'log').mockImplementation(() => {});
+// jest.spyOn(console, 'log').mockImplementation(() => {});
```
JAdshead, line 100: should this be commented out ? uncomment or remove
```suggestion
```

src/server/middleware/sendHtml.js
JAdshead: did you test this ? the purpose of this is to catch any errors that have bubbled up. its meant the final safety net. this needs to be bullet proof.

src/server/middleware/sendHtml.js
JAdshead: this is meant to be the final catch all. we can not add anything that could risk an error being thrown here.

src/server/middleware/sendHtml.js
```diff
@@ -69,30 +81,36 @@ export function renderStaticErrorPage(res) {
...
-              <div>
-                <div style="width: 70%; background-color: white; margin: 4% auto;">
-                  <h2 style="display: flex; justify-content: center; padding: 40px 15px 0px;">Loading Error</h2>
-                  <p style="display: flex; justify-content: center; padding: 10px 15px 40px;">
-                    ${message}
-                  </p>
-                </div>
+  const fallbackUrl = process.env.ONE_FALLBACK_URL;
+
+  const errorResponse = `<!DOCTYPE html>
```
JAdshead, line 86: maybe move this out of this function.

src/server/middleware/sendHtml.js
```diff
@@ -69,30 +81,36 @@ export function renderStaticErrorPage(res) {
...
-          </body>
-        </html>`);
+          </div>
+        </body>
+      </html>`;
+
+  if (fallbackUrl) {
+    await fetchCustomErrorPage(fallbackUrl, res);
+  } else {
+    safeSend(res, errorResponse);
```
JAdshead, line 112: can you you fetchCustomErrorPage to get the response and pass into safeSend here

__tests__/server/middleware/sendHtml.spec.js
JamesSingleton: @JAdshead good catch! I commented out for testing

src/server/middleware/sendHtml.js
JamesSingleton: Yeah with the other changes requested I will be moving this out now.

src/server/middleware/sendHtml.js
JamesSingleton: Any particular reason to not do it `fetchCustomErrorPage`?

src/server/middleware/sendHtml.js
JamesSingleton: I tested with responses of `application/json` and `text/html`. The ones with `application/json` threw and the ones with `text/html` did not and returned that result.

src/server/middleware/sendHtml.js
```diff
@@ -57,7 +57,27 @@ export function safeSend(res, ...payload) {
...
+  const contentType = response.headers.get('content-type');
+  const contentLength = response.headers.get('content-length');
+
+  // If the Content-Type is not text/html throw an error
+  if (!contentType.includes('text/html')) {
+    throw new Error('Content-Type was not of type text/html');
+  }
+  // If the content length is over 50kb throw an error
+  if (contentLength > 50000) {
+    throw new Error('Content-Length was over 50Kb');
```
JAdshead, line 73: maybe add a little more info about this for user in the error and does this really matter ?

src/server/middleware/sendHtml.js
```diff
@@ -57,7 +57,27 @@ export function safeSend(res, ...payload) {
   }
 }
 
-export function renderStaticErrorPage(res) {
+let errorPage;
+
+export async function fetchErrorPage(fallbackUrl) {
```
JAdshead, line 62: what about ? 
```suggestion
export async function setErrorPage(fallbackUrl) {
```

src/server/middleware/sendHtml.js
```diff
@@ -57,7 +57,27 @@ export function safeSend(res, ...payload) {
   }
 }
 
-export function renderStaticErrorPage(res) {
+let errorPage;
```
JAdshead, line 60: maybe set this to the errorResponse below. then override if updated.

src/server/middleware/sendHtml.js
```diff
@@ -69,30 +89,34 @@ export function renderStaticErrorPage(res) {
...
+              ${message}
+            </p>
+          </div>
+        </div>
+      </div>
+    </body>
+  </html>`;
+
+  if (errorPage) {
+    safeSend(res, errorPage);
```
JAdshead, line 116: if errorPage defaults to the value of errorResponse then you don't need this.

src/server/middleware/sendHtml.js
JamesSingleton: @mtomcal mentioned that we should add a limit on the size of the page. We just didn't know what size to go with so we picked 50kb

src/server/middleware/sendHtml.js
JamesSingleton: The issue would be with `message` that we change depending on the status code right?

src/server/middleware/sendHtml.js
JamesSingleton: I am ok with that suggestion

src/server/middleware/sendHtml.js
mtomcal: It rejects if it exceeds timeout. I think `fetch` needs to be `createTimeoutFetch` from the enhancers.

src/server/middleware/sendHtml.js
mtomcal: Shouldn't we prevent abuse of this API? Set a sufficiently large enough HTML file and it could go OOM for the server and sink it.

__tests__/server/utils/onModuleLoad.spec.jsx
```diff
@@ -326,6 +330,21 @@ describe('onModuleLoad', () => {
...
+      module: {
+        [CONFIGURATION_KEY]: {
+          csp,
+          errorPageUrl,
+        },
+        [META_DATA_KEY]: { version: '1.0.14' },
+      },
+      moduleName: 'some-root',
+    });
+    expect(setErrorPage).toHaveBeenCalledWith(errorPageUrl);
```
infoxicator, line 345: We probably need to look into the negative scenario where the `errorPage` doesn't load / the url is invalid or the content is too large. We don't want the app to crash if something like that happens when the rootModule is loaded, we should just add a log to warn that the error page wasn't loaded and carry on

src/server/middleware/sendHtml.js
```diff
@@ -57,7 +57,27 @@ export function safeSend(res, ...payload) {
...
+  if (!contentType.includes('text/html')) {
+    throw new Error('Content-Type was not of type text/html');
+  }
+  // If the content length is over 50kb throw an error
+  if (contentLength > 50000) {
+    throw new Error('Content-Length was over 50Kb');
+  }
+  // Read the response as text.
+  errorPage = await response.text();
+  return errorPage;
```
infoxicator, line 77: What happens when this function throws? will the root module fail to load?

We are setting the value of errorPage and then returning, is that a side effect here? @JAdshead any memory implications?

infoxicator (changes requested):
I really like that we landed on the Root module approach, as far as I understand the `setErrorPage` function will only be called once per `rootModule` load, and you can only change the custom `errorPage` by re-deploying the root module, not sure if adding that to the documentation would be useful.

src/server/middleware/sendHtml.js
JamesSingleton: The server is killed like @mtomcal and the rest of us discussed

__tests__/server/utils/onModuleLoad.spec.jsx
JamesSingleton: @infoxicator this follows everything else that fails when the root module loads. If your `csp` isn't provided we cause everything to _crash_/kill the server. @mtomcal said the same idea should be applied here that if the URL that is fetched isn't correct to kill the server. If you have a broken error page or if its too large we don't want the app starting up.

mtomcal: To add that @infoxicator, consider a new Root Module version (e.g. v2.0.0) is loaded to a running One App with a stable Root Module (e.g. v1.0.0). A Root Module v2.0.0 with an invalid error page url, will cause the Root Module v2.0.0 to not get loaded and fallback to original v1.0.0 until a fix v2.1.0 module is corrected. We do no crash the server if it exists in a previously stable state. As @JamesSingleton mentions, this follows the same pattern as how CSP behaves on loading of a new Root Module.

src/server/middleware/sendHtml.js
mtomcal: If the server is undergoing a fresh startup and an invalid Root Module is provided, then the server will fail to start up. This is the same behavior as trying to start  up a Root Module with a bad CSP configuration which results in a server exit today. See the previous comment scenario about a stable running server loading an invalid Root Module: https://github.com/americanexpress/one-app/pull/281#discussion_r481297652

src/server/middleware/sendHtml.js
mtomcal: This is still needed @JamesSingleton using `createTimeoutFetch`

src/server/middleware/sendHtml.js
mtomcal: Definitely add some context to these errors `throw new Error('[appConfig/setErrorPage] Content-Length was over 50Kb');`

mtomcal (dismissed):

__tests__/server/utils/onModuleLoad.spec.jsx
infoxicator: @mtomcal thanks for confirming, so if there is a previous working version that should be fine 👍

src/server/middleware/sendHtml.js
Francois-Esquire: Agree with @JAdshead , moving this out would make sense. Maybe in the if/else statement below where it applies? (when error page is not supplied)

Francois-Esquire (changes requested):

src/server/middleware/sendHtml.js
JamesSingleton: Ideally if they wanted a custom message they would just use a custom error page right?

src/server/middleware/sendHtml.js
JamesSingleton: I moved it into the `if/else` statement.

src/server/middleware/sendHtml.js
JamesSingleton: @mtomcal and @JAdshead I have added Tomcal's suggestion. Let me know if any more information should be added.

Francois-Esquire (dismissed):

src/server/middleware/sendHtml.js
infoxicator: Agree with @JAdshead with this small refactor if you want to apply it by just having 1 safeSend and the contents of the error change depending on the value of the `errorPageUrl` i.e. 

moving the logic of the default page to the top setting the value of errorPage to be the default, then overriding it if setErrorPage is there, and at the end do only one `safeSend(res, errorPage)` removing that if statement.

infoxicator: You are right the if statement is required to determine if sending the custom or default error page... the default error page could be moved outside and pass the message as a parameter from `renderStaticErrorPage()` but probably not worth the hassle

__tests__/integration/one-app.spec.js
```diff
@@ -862,6 +862,32 @@ describe('Tests that require Docker setup', () => {
...
+          version: '0.0.4',
+        });
+        // wait for change to be picked up
+        await waitFor(5000);
+      };
+
+      afterAll(() => {
+        writeModuleMap(originalModuleMap);
+      });
+      describe('successful fetch of error page', () => {
```
JAdshead, line 878: what about the default scenario ? when custom error is not provided ?

docs/api/modules/App-Configuration.md
```diff
@@ -454,6 +456,25 @@ The `eventLoopDelayThreshold` directive accepts a number representing the thresh
...
+
+```js
+if (!global.BROWSER) {
+  Module.appConfig = {
+    errorPageUrl: String,
+  };
+}
+```
+
+The `errorPageUrl` directive is useful for supplying a URL to a custom error page. This URL needs to return a `Content-Type` of `text/html` and a `Content-Length` of less than `50000`. The URL will get called when your root module is loaded and is rendered if and when an error occurs.
```
JAdshead, line 476: where did the `< 50000` limit come from ?

__tests__/integration/one-app.spec.js
```diff
@@ -862,6 +862,32 @@ describe('Tests that require Docker setup', () => {
       });
     });
 
+    describe('custom error page', () => {
+      const loadCustomErrorPageRoot = async () => {
+        await addModuleToModuleMap({
+          moduleName: 'frank-lloyd-root',
+          version: '0.0.4',
```
JAdshead, line 869: i don't think we needed a whole new frank for this. each new sample module version increases the cost of running integration tests. can you try to use an existing version ?

JAdshead (changes requested):

docs/api/modules/App-Configuration.md
JamesSingleton: @JAdshead it was a limit that @mtomcal and I came up with randomly I will admit. We wanted to discuss during the code review.

__tests__/integration/one-app.spec.js
JamesSingleton: I thought that was already tested here: https://github.com/americanexpress/one-app/blob/master/__tests__/integration/one-app.spec.js#L1263?

__tests__/integration/one-app.spec.js
JamesSingleton: @JAdshead we can, it was suggested to create a new one, but I can add it to an existing one if the team agrees.

docs/api/modules/App-Configuration.md
JamesSingleton: @JAdshead I changed this to be the default size of our modules

JAdshead (changes requested):
after offline discussion, warn don't throw when failing to update error page.
This is to prevent one-app failing to start when pods scale / restart due to an external factor "e.g network issue fetching error page"

__tests__/server/middleware/sendHtml.spec.js
```diff
@@ -825,7 +839,133 @@ describe('sendHtml', () => {
...
+    it('warns if url cannot be fetched', async () => {
+      const fetchError = new Error('getaddrinfo ENOTFOUND');
+      global.fetch = jest.fn(() => Promise.reject(fetchError));
+
+      await setErrorPage(errorPageUrl);
+      expect(console.warn).toHaveBeenCalledTimes(1);
+      expect(console.warn).toHaveBeenCalledWith('Could not fetch the URL', fetchError);
+    });
+
+    it('warns if content-length is greater than 50Kb', async () => {
```
JAdshead, line 954:
```suggestion
    it('warns if content-length is greater than 244kb', async () => {
```

docs/api/modules/App-Configuration.md
JAdshead, line 476:
```suggestion
The `errorPageUrl` directive is useful for supplying a URL to a custom error page. This URL should return a `Content-Type` of `text/html` and a `Content-Length` of less than `244000`. The URL will get called when your root module is loaded and is rendered if and when an error occurs. It is recommended that you keep the custom error page as small as possible.
```

prod-sample/sample-modules/frank-lloyd-root/0.0.0/src/config.js
```diff
@@ -56,4 +56,5 @@ export default {
     ],
   },
   createSsrFetch: createFrankLikeFetch,
+  errorPageUrl: 'https://sample-cdn.frank/error.html',
```
JAdshead, line 59: is this going to conflict with the other error page test ?

src/server/middleware/sendHtml.js
```diff
@@ -57,42 +58,70 @@ export function safeSend(res, ...payload) {
...
-export function renderStaticErrorPage(res) {
+let errorPage;
+
+export async function setErrorPage(fallbackUrl) {
+  try {
+    const timeoutFetch = createTimeoutFetch(6e3)(fetch);
+    const response = await timeoutFetch(fallbackUrl);
+    const contentType = response.headers.get('content-type');
+    const contentLength = response.headers.get('content-length');
+
```
JAdshead, line 69: maybe check the fetch call was successful ?

src/server/middleware/sendHtml.js
```diff
@@ -57,42 +58,70 @@ export function safeSend(res, ...payload) {
...
+export async function renderStaticErrorPage(res) {
   if (!res.statusCode) {
     res.status(500);
   }
   console.info(`renderStaticErrorPage status ${res.statusCode}`);
 
-  let message = 'Sorry, we are unable to load this page at this time. Please try again later.';
-  if (res.statusCode >= 400 && res.statusCode < 500 && res.statusCode !== 404) {
+  if (errorPage) {
+    console.log('INSIDE IF');
```
JAdshead, line 94:
```suggestion
```

JAdshead (changes requested):

prod-sample/sample-modules/frank-lloyd-root/0.0.0/src/config.js
JamesSingleton: It shouldn't

JamesSingleton: Turns out it did, I resolved it.

src/server/middleware/sendHtml.js
infoxicator: For reference: After some discussions with the team we decided that killing the server on an invalid custom error page is not ideal. Unlike the CSP the custom error page functionality is not critical for the server to work correctly since we can always fallback to the static error page as before. furthermore the custom error page is loaded from an external source which might not be reliable at some point when the server is trying to restart (due to autoscaling or pod restart)

__tests__/server/middleware/sendHtml.spec.js
```diff
@@ -825,7 +839,133 @@ describe('sendHtml', () => {
...
+      global.fetch = jest.fn(() => Promise.reject(fetchError));
+
+      await setErrorPage(errorPageUrl);
+      renderStaticErrorPage(res);
+
+      expect(res.send).toHaveBeenCalledTimes(1);
+      expect(res.send.mock.calls[0][0]).toContain('<!DOCTYPE html>');
+      expect(res.send.mock.calls[0][0]).toContain('<meta name="application-name" content="one-app">');
+    });
+    it('returns provided error page if provided', async () => {
```
infoxicator, line 854:
```suggestion
    it('returns a custom error page if provided', async () => {
```

__tests__/integration/one-app.spec.js
```diff
@@ -862,6 +862,32 @@ describe('Tests that require Docker setup', () => {
...
+        // wait for change to be picked up
+        await waitFor(5000);
+      };
+
+      afterAll(() => {
+        writeModuleMap(originalModuleMap);
+      });
+      describe('successful fetch of error page', () => {
+        beforeAll(loadCustomErrorPageRoot);
+        test('responses with custom error page', async () => {
```
infoxicator, line 880:
```suggestion
        test('responds with a custom error page', async () => {
```

__tests__/server/middleware/sendHtml.spec.js
```diff
@@ -825,7 +839,133 @@ describe('sendHtml', () => {
...
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+  </head>
+  <body>
+    <div>
+      <h1>Custom Error Page</h1>
+      <p>This is a custom error page.</p>
+    </div>
+  </body>
+  </html>`;
+    it('fetches errorPageUrl', async () => {
```
infoxicator, line 914: This tests it is very similar to `returns provided error page if provided` above. I know we are testing the `setErrorPage` function and the `renderStaticErrorPage` separately but maybe gather some more feedback from the team to see if it is required

infoxicator (approved):
Apart from a couple of optionals `suggestions` to the name of the tests this looks really good 👍

__tests__/server/middleware/sendHtml.spec.js
JamesSingleton: So this tests `setErrorPage` which gets called during `onModuleLoad`, the other test is testing that `renderStaticErrorPage` is actually sending that custom error page. I believe there should be two separate tests.

infoxicator (approved):

Francois-Esquire (dismissed):

src/server/middleware/sendHtml.js
JamesSingleton: @JAdshead if it isn't a 200 response, what do you suggest that we do?

JAdshead: this needs to handle that scenario, you don't want to be setting the domains 404/500 error page to be this error page. `response.text()` won't throw for those scenarios and if it does the error will be misleading

JamesSingleton: Fixed in https://github.com/americanexpress/one-app/pull/281/commits/5c5d4dabcee6e57da852f32b2eb4b7bc519f9d89

src/server/middleware/sendHtml.js
```diff
@@ -57,42 +58,74 @@ export function safeSend(res, ...payload) {
...
+    // Warn if the Content-Type is not text/html
+    if (!contentType.includes('text/html')) {
+      console.warn('[appConfig/errorPageUrl] Content-Type was not of type text/html and may not render correctly');
+    }
+    // Warn if the content length is over 244kb
+    if (contentLength > 250e3) {
+      console.warn('[appConfig/errorPageUrl] Content-Length is over 244Kb and may have an impact on performance');
+    }
+
+    if (response.status !== 200) {
```
JAdshead, line 79: this should be the first check.

src/server/middleware/sendHtml.js
```diff
@@ -57,42 +58,74 @@ export function safeSend(res, ...payload) {
...
+      console.warn('Custom error page did not return a status of 200... Falling back to default error page');
```
JAdshead, line 80:
```suggestion
      console.warn('Failed to fetch custom error page with status:', response.status);
```

JAdshead (changes requested):

Francois-Esquire (approved):

mtomcal (approved):

JAdshead (approved):

PR #286, opened 2020-08-26
fix(abort-controller): add ie11 polyfil

JAdshead:
## Description

Polyfil Abort Controller for ie 11 support.

## Motivation and Context

Reports of bugs in ie11 related to AbortController not being available 

## How Has This Been Tested?

Tested for regression locally. Confirmed fixes reported errors in ie11

## Types of Changes

- [ ] Dependency update

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "@americanexpress/one-app-ducks": "^4.1.1",
     "@americanexpress/one-app-router": "^1.1.0",
     "@americanexpress/vitruvius": "^2.0.0",
+    "abort-controller": "^3.0.0",
     "body-parser": "^1.19.0",
     "chalk": "^4.1.0",
     "compression": "^1.7.4",
--- a/src/client/client.js
+++ b/src/client/client.js
@@ -18,6 +18,9 @@
  * Main App
  * Set up the global store and kick off rendering
  */
+
+// polyfill for ie 11
+import 'abort-controller/polyfill';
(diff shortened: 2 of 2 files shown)


src/client/client.js
```diff
@@ -18,6 +18,9 @@
  * Main App
  * Set up the global store and kick off rendering
  */
+
+// polyfill for ie 11
+import 'abort-controller/polyfill';
```
anescobar1991, line 23: does this automatically only get included in the legacy bundle?

anescobar1991:
Thanks @JAdshead!

package.json
```diff
@@ -83,6 +83,7 @@
     "@americanexpress/one-app-ducks": "^4.1.1",
     "@americanexpress/one-app-router": "^1.1.0",
     "@americanexpress/vitruvius": "^2.0.0",
+    "abort-controller": "^3.0.0",
```
anescobar1991, line 86: no package-lock.json change?

JAdshead: its already provided by fetch-enhancers and was deduped so no change

infoxicator (approved):

src/client/client.js
JamesSingleton: @anescobar1991 yeah it only works on branches from the repo itself, not on forks.

JAdshead:
declining in favour of - https://github.com/americanexpress/one-app/pull/289

PR #301, opened 2020-09-10
Docs/partial rendering

Francois-Esquire:
Recipe/guide for Partial Rendering with sample walk through.

## Types of Changes

- [ ] Dependency update

## Checklist:

Docs to using partial rendering.


docs/recipes/Partial-Rendering.md
```diff
@@ -4,18 +4,172 @@
...
+
+// ...
+```
+
+Since partial rendering is configurable only during server side rendering,
+a partial render can source props from the request body via initial state building.
+The Holocron config adds a reducer which stores the props from the request body.
+
+```js
 // ...
```
JAdshead, line 95: I'm not sure if its clear enough that this is a continuation.

```suggestion
```js
// continued RootModule

```

docs/recipes/Partial-Rendering.md
```diff
@@ -4,18 +4,172 @@
...
+  const { params: { moduleName } } = props;
+  return (
+    <RenderModule
+      moduleName={moduleName}
+      props={props}
+    />
+  );
+}
+
+export default function createPartialRoute(getPartialRenderProps) {
```
JAdshead, line 61: should we start with a simpler example of a plain `ModuleRoute` and then suggest this as a possibility if they are creating multiple partial routes.

JAdshead (changes requested):

docs/recipes/Partial-Rendering.md
Francois-Esquire: Good suggestion, I'll update with this

docs/recipes/Partial-Rendering.md
Francois-Esquire: I agree with this point, I'll update

infoxicator (commented):
I think we should maybe focus this example on a real use case like rendering an email? and also expand on the need for `<dangerously-return-only-doctype />` and also an example of why that is required?

docs/recipes/Partial-Rendering.md
```diff
@@ -4,18 +4,172 @@
 
 # Partial Rendering
 
+In this recipe, we will go over the setup required to start using partial rendering with One App,
```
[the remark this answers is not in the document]
Francois-Esquire: I've updated this a bit but I've kept the suggested sentence (5809765)

Francois-Esquire:
> I think we should maybe focus this example on a real use case like rendering an email? and also expand on the need for `<dangerously-return-only-doctype />` and also an example of why that is required?

@infoxicator I've updated to include an email example using `dangerously-return-only-doctype` (5809765)

JAdshead (dismissed):

docs/recipes/Partial-Rendering.md
```diff
@@ -4,24 +4,324 @@
...
+  return (
+    <dangerously-return-only-doctype
+      dangerouslySetInnerHTML={{ __html: html }}
+    />
+  );
+}
+```
+
+Do keep in mind that partial rendering is meant to be performed server-side, we should not attempt
+this example above in the browser in normal circumstances.
```
mtomcal, line 309:
```suggestion
this example above in the browser under normal circumstances.
```

mtomcal (changes requested):
Small nit. Overall quite well written.

mtomcal (approved):

JAdshead (approved):

PR #306, opened 2020-09-16
fix(watchLocalModules): fix watcher crashing due to build errors

JamesSingleton:
This was caused when a module would throw a linting error or something similar
during the building of the module. This was not being caught and therefore was causing
the app to crash.

## Description

Fixes an issue where a build failure when doing `npm run watch:build` in a module would cause the entire app to crash.

## Motivation and Context

Improve DX.

## How Has This Been Tested?

Locally

## Types of Changes

- [ ] Dependency update

## Checklist:

N/A other than better DX


src/server/utils/watchLocalModules.js
```diff
@@ -39,28 +39,32 @@ export default function watchLocalModules() {
...
+        onModuleLoad
+      ));
 
-    const modules = getModules().set(moduleNameChangeDetectedIn, module);
-    resetModuleRegistry(modules, getModuleMap());
+      const modules = getModules().set(moduleNameChangeDetectedIn, module);
+      resetModuleRegistry(modules, getModuleMap());
+    } catch (error) {
+      console.error(error);
+    }
```
mtomcal, line 68: This `catch` was the only code change aside from formatting above?

JAdshead (approved):

src/server/utils/watchLocalModules.js
JamesSingleton: This was the only change and updating `holocron`

mtomcal (approved):

Francois-Esquire (approved):

nellyk (approved):

PR #313, opened 2020-09-23
feat(ssrServer): enable urlencoded post data

sumanramagoni:
post parameters are available in `req.body` only when request `content-type` is `application/json`. There is no way to handle default encoding method.

## Description
Feature to enable form post with default encoding method `application/x-www-form-urlencoded`

## Motivation and Context
There is a need for the clients to use form post for sharing information on to the application. This feature is already available in lower version of one-app.

## How Has This Been Tested?
Tested it locally by posting parameters, we could verify all of the post parameters in `req.body`

## Types of Changes

- [ ] Dependency update

## Checklist:

NA

### DIFF
--- a/src/server/ssrServer.js
+++ b/src/server/ssrServer.js
@@ -23,7 +23,7 @@ import path from 'path';
 import express from 'express';
 import compression from 'compression';
 import cookieParser from 'cookie-parser';
-import { json } from 'body-parser';
+import { json, urlencoded } from 'body-parser';
 import helmet from 'helmet';
 import cors from 'cors';
 
@@ -96,6 +96,7 @@ export function createApp({ enablePostToModuleRoutes = false } = {}) {
       '*',
       addSecurityHeaders,
       json({ limit: '0kb' }), // there should be no body
+      urlencoded({ limit: '0kb' }), // there should be no body
       cors({ origin: false }) // disable CORS
     );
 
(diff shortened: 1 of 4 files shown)


src/server/ssrServer.js
```diff
@@ -96,13 +96,15 @@ export function createApp({ enablePostToModuleRoutes = false } = {}) {
...
       json({ limit: '0kb' }), // there should be no body
+      urlencoded({ limit: '0kb' }), // there should be no body
       cors({ origin: false }) // disable CORS
     );
 
     app.post(
       '*',
       addSecurityHeaders,
       json({ limit: '15kb' }),
+      urlencoded({ limit: '15kb' }),
```
Francois-Esquire, line 107: If we are okay with the limit and feel secure about supporting this, I think this would be a good feature to bring in!
@sumanramagoni can you please add documentation and I think this is a feature, not a fix btw.

Francois-Esquire (changes requested):
I have a small list of requests:

- Docs need to be added

then its good by me!

Possibly do we want to make this an opt in? users can set to true to use

Francois-Esquire:
Btw, thank you very much for your contribution 🙏

src/server/ssrServer.js
sumanramagoni: yea, I've realized at later point in time that this should be considered as feature. 
shall I close this pr and raise new one with feature?

nellyk: you can edit the pull request and update the commit message, no need to close the PR

sumanramagoni: done.
I added documentation as well.

sumanramagoni:
> I have a small list of requests:
> 
> * Docs need to be added
> 
> then its good by me!
> 
> Possibly do we want to make this an opt in? users can set to true to use

I've added documentation under `Making-An-Api-Call`

Francois-Esquire (approved):

JAdshead (approved):

PR #338, opened 2020-10-07
Docs/server side rendering

JamesSingleton:
## Types of Changes

- [ ] Dependency update

## Checklist:

### DIFF
--- a/docs/recipes/Enabling-Serverside-Rendering.md
+++ b/docs/recipes/Enabling-Serverside-Rendering.md
@@ -2,8 +2,103 @@
 [👈 Return to Overview](./README.md)
 <!--ONE-DOCS-HIDE end-->
 
-# Enabling Serverside Rendering
+# Enabling Server-side Rendering
 
-Recipe is forthcoming.
+One App will always attempt render your module on the server before sending html back to the client. However, it won't wait for any asynchronous tasks, such as data fetching, to complete before doing so.
+
+The following options provide ways to ensure that server side requested data is included as part of the initial render.
+
+## Using `loadModuleData`
+
+One App v5 introduced [`loadModuleData`](https://one-amex-docs.americanexpress.com/en-us/one-app/api/modules/loading-data) which is invoked on the server before rendering a module.
+
+```jsx
+YourModule.holocron = {
+  name: 'your-module',
+  // Runs on both Server and Browser
+  loadModuleData: async ({ store, fetchClient, ownProps }) => {
+    store.dispatch({ type: 'LOADING_API' });
+    const response = await fetchClient('https://api.example.com', ownProps.options);
+    const data = await response.json();
+    store.dispatch({ type: 'LOADED_API', data });
+  },
+};
+```
+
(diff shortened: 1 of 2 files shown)


infoxicator:
@JamesSingleton are we still having a separate guide on `how to make API calls`? this pretty much sums most of it

JamesSingleton:
> @JamesSingleton are we still having a separate guide on `how to make API calls`? this pretty much sums most of it

@infoxicator 🤔, we can remove it if you think it no longer needs to be there.

infoxicator:
> > @JamesSingleton are we still having a separate guide on `how to make API calls`? this pretty much sums most of it
> 
> @infoxicator 🤔, we can remove it if you think it no longer needs to be there.

I think we could remove it or... convert it into a specific guide for `fetchye` to include client side config and to explain why `fetchye` is useful in the one app world (caching between modules)... maybe a new ticket for it 👍

docs/recipes/Enabling-Serverside-Rendering.md
```diff
@@ -2,8 +2,103 @@
...
+# Enabling Server-side Rendering
 
-Recipe is forthcoming.
+One App will always attempt render your module on the server before sending html back to the client. However, it won't wait for any asynchronous tasks, such as data fetching, to complete before doing so.
+
+The following options provide ways to ensure that server side requested data is included as part of the initial render.
+
+## Using `loadModuleData`
+
+One App v5 introduced [`loadModuleData`](https://one-amex-docs.americanexpress.com/en-us/one-app/api/modules/loading-data) which is invoked on the server before rendering a module.
```
infoxicator, line 13: might be worth mentioning that `loadModuleData` runs on the server AND client again here? I see the comment below but might be important to clarify since it could cause confusion

docs/recipes/Enabling-Serverside-Rendering.md
```diff
@@ -2,8 +2,103 @@
...
+  return (
+    <Fragment>
+      {/* Render data */}
+    </Fragment>
+  );
+};
+
+BookList.holocron = {
+  loadModuleData: async ({ store: { dispatch, getState }, fetchClient }) => {
+    if (global.BROWSER) {
```
infoxicator, line 51: Did we find out which one includes / removes the code from the client bundle? `global.BROWSER` or `!global.BROWSER`

infoxicator (dismissed):
Optional extra clarification about running both client and server side but not compulsory to add it

docs/recipes/Enabling-Serverside-Rendering.md
JamesSingleton: So `if (!global.BROWSER)` would prevent it from being bundled in the client code

infoxicator (approved):

JamesSingleton:
In response to @infoxicator comment about removing the fetching data, I think we should keep it so we can other examples like using `react-query` and other libraries.

JAdshead (approved):

PR #354, opened 2020-10-20
docs(articles): getting started

Francois-Esquire:
Getting started guide for beginners. The article connects the dots between the ecosystem and the runtime.

## Description

The article starts with creating a module and moves on to other topics with One App and Holocron.

## Types of Changes

- [ ] Dependency update

## Checklist:

Documentation


docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+npm run watch:build
+```
+
+and this will watch for any changes made to your module, then update `one-app-runner`
+with the rebuilt module bundle. The command above uses `one-app-bundler` which can
+also be used to bundle our Holocron module.
+
+```bash
+npm run build
+```
```
JAdshead, line 81:
```suggestion
You can use also build the module outside of watch mode:`

```bash
npm run build
```
```

**Review by JAdshead — CHANGES_REQUESTED** [review]
Good start, i think we might want to focus on the micro-front end part then reference the recipes for everything else such as styling.

File: docs/articles/Getting-Started.md [hunk 2]
```diff
@@ -0,0 +1,387 @@
...
+### Generating a Module
+
+The first step to get started with One App is to generate a Holocron module.
+There are two different Holocron module types, root and child module. Let us
+start by creating a root Holocron module, the entry point to our micro frontend.
+
+```bash
+export NODE_ENV=development
+
+npx -p yo -p @americanexpress/generator-one-app-module \\
```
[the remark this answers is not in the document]
Francois-Esquire: It's a pretty long command and breaks flex positioning in mobile devices.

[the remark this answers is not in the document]
JamesSingleton: 🤔 but when it gets put to the docs it isn't a simple copy and paste

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+
+This guide will get us started with One App and will go over some of the
+fundamental concepts of the framework. We will start from the beginning
+by creating our first Holocron module, then we will go through a few
+One App essentials to get up and running quickly.
+
+## 📖 Table of Contents
+* [Generating a Module](#generating-a-module)
+* [Running One App](#running-one-app)
+* [Adding CSS Styles](#adding-css-styles)
```
nellyk, line 21: i really like how you've broken this down into different sections

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+* [Creating Routes](#creating-routes)
+* [Module State and Data](#module-state-and-data)
+* [Configuring One App](#configuring-one-app)
+* [Development](#development)
```
nellyk, line 25: Maybe we should be more specific about this, what do you think? `Setting up your development environment`

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,387 @@
...
+* [Adding CSS Styles](#adding-css-styles)
+* [Creating Routes](#creating-routes)
+* [Module State & Data](#module-state-&-data)
+* [Configuring One App](#configuring-one-app)
+* [Development](#development)
+
+### Generating a Module
+
+The first step to get started with One App is to generate a Holocron module.
+There are two different Holocron module types, root and child module. Let us
```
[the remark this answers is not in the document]
nellyk: i'd love if we could add images to describe this

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,387 @@
...
+
+npx -p yo -p @americanexpress/generator-one-app-module \\
+-- yo @americanexpress/one-app-module
+```
+
+Once the command is executed, you will be prompted to fill out a few questions
+about your new module before it is generated. Starting off, we will want to name
+our module and make sure that it is a root module; feel free to answer as you
+wish for the other questions. Once the root module is generated we will be able
+to start developing with One App.
```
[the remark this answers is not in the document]
nellyk: A gif would be amazing to show this

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+>
+> [`react-dom`](https://reactjs.org/docs/react-dom.html)
+>
+> [`holocron`](https://github.com/americanexpress/holocron/tree/main/packages/holocron)
+
+### Running One App
+
+Every Holocron module that is generated comes with pre-installed scripts that can
+be run afterwards. One of these scripts is the `npm start` script that starts up
+`one-app-runner`, which is our primary tool for local development.
```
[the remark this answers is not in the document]
nellyk: we could also link to other commands that are available

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+
+and this will watch for any changes made to your module, then update `one-app-runner`
+with the rebuilt module bundle. The command above uses `one-app-bundler` which can
+also be used to bundle our Holocron module.
+
+```bash
+npm run build
+```
+
+When ready to publish to production, make sure to set `NODE_ENV=production` before building.
```
nellyk, line 83: should we mention why here?

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+Inside the `package.json` of your Holocron module, you can
+include a `"one-amex": {}` property to configure some of the
+tools in our ecosystem.
+
+```json
+{
+  "name": "my-module",
+  "one-amex": {
+    "bundler": {
+      "webpackConfigPath": "webpack.config.js"
```
nellyk, line 96: Should we recommend  `webpackConfigPath` for the first setup

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+
+`src/components/MyModule.jsx`
+
+```jsx
+import React from 'react';
+import { useSelector } from 'react-redux';
+
+export default function MyModule() {
+  const { theme } = useSelector(
+    (state) => state.get('config'),
```
infoxicator, line 338: maybe consider this example to load a single config value from the state? for best practices and performance we should recommend avoiding `state.ToJS()` if possible

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+    (state) => state.toJS()
+  );
```
infoxicator, line 340:
```suggestion
 const theme = useSelector((state) => state.getIn(['config', 'theme']));
```

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+One App comes with a security standard in practice, progressive web app capability and
+many other features, its feature set can be configured by Holocron modules to your
+exact specification.
+
+This guide will get us started with One App and will go over some of the
+fundamental concepts of the framework. We will start from the beginning
+by creating our first Holocron module, then we will go through a few
+One App essentials to get up and running quickly.
+
+## 📖 Table of Contents
```
infoxicator, line 18: I have seen the CSP is a common pitfall...

What about adding a section for a basic CSP configuration at least explaining that one-app **Blocks** everything by default and that you won't be able to make any api calls unless you configure your CSP.

This should be explained as one of the main advantages of using one app as a best practice security feature.

infoxicator (changes requested):

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+
+function DataVisualizer({ data, loaded }) {
+  return (
+    <div>
+      {/* do something with the data when it loads */}
+    </div>
+  );
+}
+
+export default function MyModule({ moduleState = {}, moduleLoadStatus = 'loading' }) {
```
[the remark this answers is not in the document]
nellyk: i agree on this

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+> **Recipes**
+>
+> [Adding Styles Recipe](../recipes/adding-styles)
+
+### Creating Routes
+
+One App has built in dynamic routing that uses each Holocron module to
+build out the router. There is a special property that we can assign to
+our module `Module.childRoutes` which would allow us to add routes to
+One App. `childRoutes` should be assigned a function that is given
```
10xLaCroixDrinker, line 181: It only needs to be a function when you need to use the store for something, otherwise it can be an array or just a route with or without nested routes

10xLaCroixDrinker (changes requested):

docs/articles/Getting-Started.md
Francois-Esquire: I agree with you however CSP is an advanced topic that deserves an article or a set of reference documentation on its own. Another concern with adding a section is it starts to make the document heavy for a getting started guide, no?

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
+# Getting Started
+
+One App is a web application that combines together multiple experiences into a single,
+cohesive runtime. These experiences, or Holocron modules as we call them empower teams
```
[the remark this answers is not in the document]
Francois-Esquire: Agree with you, I'll update

docs/articles/Getting-Started.md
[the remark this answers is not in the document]
Francois-Esquire: If putting it on one line would make it less confusing, I'm all for it. Will update

docs/articles/Getting-Started.md
Francois-Esquire: Good point on that, I wanted readers to know that it is possible, however as you both mentioned, it is an advanced feature.

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+}
+```
+
+> #### More Info
+>
+> **Recipes**
+>
+> [Adding Styles Recipe](../recipes/adding-styles)
+
+### Creating Routes
```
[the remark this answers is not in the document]
Francois-Esquire: I actually did want to talk about `RenderModule`! The reason I left it out was the size of the article was growing... Inthe future, we should think about making a recipe on the different ways to render a module.

docs/articles/Getting-Started.md
Francois-Esquire: TIL, thank you for sharing that 😄

docs/articles/Getting-Started.md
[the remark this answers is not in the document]
Francois-Esquire: My only concern with adding other commands is the cognitive overload... I'd like to keep as simple as possible, and adding the links gives people the extra docs they need if a package captures their attention.

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+
+```bash
+npm start
+```
+
+When `one-app-runner` is fully loaded and running, we can navigate to `http://localhost:3000`
+and view our Holocron module in the browser. In another terminal window you can run
+
+```bash
+npm run watch:build
```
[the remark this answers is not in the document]
Francois-Esquire: Good. suggestion, I will be adding this manually.

docs/articles/Getting-Started.md
Francois-Esquire: I will add this manually.

docs/articles/Getting-Started.md
Francois-Esquire: Good points, I will update

docs/articles/Getting-Started.md
```diff
@@ -0,0 +1,386 @@
...
+  &:focus,
+  &:hover {
+    background-color: blue;
+  }
+}
+```
+
+We can import the stylesheet into our module and use CSS modules to access the class name
+for each selector by its name. The class names are unique when they are generated which
+avoids unwanted cascading of styles in our document.
```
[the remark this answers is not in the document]
Francois-Esquire: I will add a bit to this sentence

docs/articles/Getting-Started.md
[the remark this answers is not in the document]
Francois-Esquire: Fair point, will update this

docs/articles/Getting-Started.md
Francois-Esquire: I totally agree that this section is bare compared to the rest, I'll update it

docs/articles/Getting-Started.md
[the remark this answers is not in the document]
Francois-Esquire: About ordering sections, I get where you are coming from when switching the two sections.. the reason I'd want to keep the CSS as the first activity in the module is the ease to set up, a core part of web development (CSS) and direct visual feedback will give readers (especially new to One App or JavaScript) a soft start for the first thing they do in a module. I feel that would segway into more JavaScript heavy sections that follow.

infoxicator (approved):

docs/articles/Getting-Started.md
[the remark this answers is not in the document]
nellyk: agree with your take @Francois-Esquire  since users can still load the webpage from the root, afterwards they can look at adding routes

docs/articles/Getting-Started.md
nellyk: yeah we should add it to separate PR and maybe mention the article here once we have the CSP content

nellyk (approved):

docs/articles/Getting-Started.md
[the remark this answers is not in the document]
Francois-Esquire: We can plan to add images later,  I agree would be so helpful!

docs/articles/Getting-Started.md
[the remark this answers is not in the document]
Francois-Esquire: Updated in 789f657

docs/articles/Getting-Started.md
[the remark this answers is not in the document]
Francois-Esquire: updated 789f657

docs/articles/Getting-Started.md
Francois-Esquire: I've added these

docs/articles/Getting-Started.md
Francois-Esquire: Thank you!

docs/articles/Getting-Started.md
Francois-Esquire: Updated

docs/articles/Getting-Started.md
Francois-Esquire: Updated

JamesSingleton (approved):

JAdshead (approved):

10xLaCroixDrinker (approved):

PR #356, opened 2020-10-21
fix(security): specify min TLS version

JamesSingleton:
## Description

Specify min TLS version

## Types of Changes

- [ ] Dependency update

## Checklist:

### DIFF
--- a/src/server/listen.js
+++ b/src/server/listen.js
@@ -42,11 +42,9 @@ function listenHttps(app, cb) {
   const serverOptions = {
     key: fs.readFileSync(process.env.HTTPS_PRIVATE_KEY_PATH),
     cert: fs.readFileSync(process.env.HTTPS_PUBLIC_CERT_CHAIN_PATH),
+    minVersion: 'TLSv1.2',
   };
 
-  // TODO: support all HTTPS/TLS options
-  // https://nodejs.org/api/https.html#https_https_createserver_options_requestlistener
-  // https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options
   if (process.env.HTTPS_TRUSTED_CA_PATH) {
     Object.assign(serverOptions, {
       // self-signed certificate to the upstream load balancer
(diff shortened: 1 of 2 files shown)


src/server/listen.js
```diff
@@ -42,6 +42,7 @@ function listenHttps(app, cb) {
   const serverOptions = {
     key: fs.readFileSync(process.env.HTTPS_PRIVATE_KEY_PATH),
     cert: fs.readFileSync(process.env.HTTPS_PUBLIC_CERT_CHAIN_PATH),
+    minVersion: 'TLSv1.2',
   };
 
   // TODO: support all HTTPS/TLS options
```
nellyk, line 48: Should we remove this?

src/server/listen.js
```diff
@@ -42,6 +42,7 @@ function listenHttps(app, cb) {
   const serverOptions = {
     key: fs.readFileSync(process.env.HTTPS_PRIVATE_KEY_PATH),
     cert: fs.readFileSync(process.env.HTTPS_PUBLIC_CERT_CHAIN_PATH),
+    minVersion: 'TLSv1.2',
```
nellyk, line 45: if you have any version greater than `TLSv1.2`  will that be supported?

JAdshead: could this be considered breaking ?

JAdshead (approved):

JAdshead:
I think we should consider this as a security fix rather than a feature

src/server/listen.js
JamesSingleton: @nellyk this is setting the min version https://nodejs.org/dist/latest-v12.x/docs/api/tls.html#tls_tls_createsecurecontext_options

nellyk: thank you, that helped 😄  , could be a breaking change though?
> It is not recommended to use less than TLSv1.2, but it may be required for interoperability.

JamesSingleton: @10xLaCroixDrinker said that we would not be considering this a breaking change

infoxicator (approved):

JAdshead (approved):

PR #369, opened 2020-11-04
test(integration): remove surge dependency

infoxicator:
Remove surge from the dependency list since is not needed.

## Description

- During integration tests we are using `npx surge` so the surge dependency is not required by one-app.
- By removing `surge` as a dependency we also remove the dependency on `tarr` a package created by the surge-cli developers and not the officially supported `tar`
- Add missing env Var `'${{ secrets.HEROKU_APP_URL }}'` to the `CORS` configuration section instead of the hardcoded domain.

## Motivation and Context

Removes unnecessary dependency

## How Has This Been Tested?

Successful integration test run using the new configuration: 
https://github.com/infoxicator/one-app/runs/1353253663?check_suite_focus=true

## Types of Changes

- [x] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/.github/workflows/integrations_test.yml
+++ b/.github/workflows/integrations_test.yml
@@ -29,7 +29,7 @@ jobs:
          SURGE_TOKEN: ${{ secrets.SURGE_TOKEN }}
         run:  |
          npm run build:sample-modules -- --archive-built-artifacts --bundle-statics-origin=${{ secrets.SURGE_DOMAIN }}
-         echo 'https://one-app-sample.herokuapp.com' >> sample-module-bundles/CORS
+         echo '${{ secrets.HEROKU_APP_URL }}' >> sample-module-bundles/CORS
       - name: Teardown Surge
         env:
          SURGE_LOGIN: ${{ secrets.SURGE_LOGIN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/integrations_test.yml
```diff
@@ -29,7 +29,7 @@ jobs:
          SURGE_TOKEN: ${{ secrets.SURGE_TOKEN }}
         run:  |
          npm run build:sample-modules -- --archive-built-artifacts --bundle-statics-origin=${{ secrets.SURGE_DOMAIN }}
-         echo 'https://one-app-sample.herokuapp.com' >> sample-module-bundles/CORS
+         echo '${{ secrets.HEROKU_APP_URL }}' >> sample-module-bundles/CORS
```
nellyk, line 32: good catch

nellyk (approved):

nellyk:
if you do update this workflow https://github.com/americanexpress/one-app/actions?query=workflow%3A%22One+App+Integration+Tests%22 we can test this would work with the integration tests

JAdshead (approved):

Francois-Esquire (approved):

PR #373, opened 2020-11-04
docs(recipes): monitoring one app

JAdshead:
## Description

Basic info on metrics and monitoring.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/docs/recipes/Monitoring-One-App.md
+++ b/docs/recipes/Monitoring-One-App.md
@@ -4,6 +4,41 @@
 
 # Monitoring One App
 
-Recipe is forthcoming.
+One App includes a metrics server which exposes a set of metrics using the [prom-client](https://github.com/siimon/prom-client)
+which can be easily scraped by a service such as [Prometheus](https://prometheus.io/docs/introduction/overview/).
+Metrics will be exposed on the port configured by the environment variable `HTTP_METRICS_PORT`.
+They can collected from the `/metrics` path, for example `http://localhost:3005/metrics`.
+
+> HTTP_METRICS_PORT defaults to `3005`.
+
+One App makes the following metrics available:
+
+- One App Version
+- Circuit breaker, performance including open, close events,
+- Module Map, updates, restarts and errors
+- CPU
+- Memory
+- Garbage collection
+- Eventloop lag
+
+
+## Grafana
+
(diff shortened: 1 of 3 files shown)


nellyk (dismissed):

docs/recipes/Monitoring-One-App.md
```diff
@@ -4,6 +4,41 @@
 
 # Monitoring One App
 
-Recipe is forthcoming.
+One App includes a metrics server which exposes a set of metrics using the [prom-client](https://github.com/siimon/prom-client)
+which can be easily scraped by a service such as [Prometheus](https://prometheus.io/docs/introduction/overview/).
+Metrics will be exposed on the port configured by the environment variable `HTTP_METRICS_PORT`.
+They can collected from the `/metrics` path, for example `http://localhost:3005/metrics`.
```
infoxicator, line 10:
```suggestion
They can be collected from the `/metrics` path, for example `http://localhost:3005/metrics`.
```

docs/recipes/Monitoring-One-App.md
```diff
@@ -4,6 +4,41 @@
...
+## Monitoring Locally
+
+It possible to run Prometheus and Grafana against a locally running build of one-app.
+While not an ideal performance indicator you could use it to gauge the impact your module might have.
+
+Run the following inside of One App to start the performance monitoring locally:
+
+`npm run monitor:performance`
+
+> Performance monitoring will not run with one-app runner.
```
infoxicator, line 39: do we have agreement on the spelling of one-app-runner? always seen it with 2 dashes.

Also what's the reason it doesn't work with it? is it because the docker port is not exposed?

infoxicator (changes requested):

docs/recipes/Monitoring-One-App.md
JAdshead, line 39:
```suggestion
> Performance monitoring will not run with one-app-runner.
```

JAdshead: i think it should be `one-app-runner`.

It could work but you would need to clone one-app and then run `npm run monitor:performance` from inside of one-app and then start your module. this seems like a complicated way of doing it. we could add an option to one-app-runner to start with the monitoring but i don't think your getting much when doing so.

infoxicator (approved):

JamesSingleton (approved):

PR #387, opened 2020-12-01
docs(recipes/making-an-api-call): update

JAdshead:
## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/docs/recipes/Making-An-Api-Call.md
+++ b/docs/recipes/Making-An-Api-Call.md
@@ -9,7 +9,7 @@
 Making API calls within a One App module has some additional considerations over a
 traditional client side React application.
 
-A basic data fetching example in a client side React app might look like the following:
+A basic data fetching example in a client side React app using the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) might look like the following:
 
 ```jsx
 const BooksModule = () => {
--- a/docs/recipes/Making-An-Api-Call.md
+++ b/docs/recipes/Making-An-Api-Call.md
@@ -13,27 +13,27 @@ A basic data fetching example in a client side React app might look like the fol
 
 ```jsx
 const BooksModule = () => {
-  const [{ Books, isLoading, fetchError }, setData] = useState({});
+  const [{ books, isLoading, fetchError }, setData] = useState({});
 
   useEffect(() => {
     const fetchBooks = async () => {
       try {
-        setData({ Books, isLoading: true, fetchError: false });
+        setData({ books, isLoading: true, fetchError: false });
         const response = await fetch('https://some-data-server.com/api/v1/books');
         if (response.ok) {
           const newBooks = await response.json();
-          setData({ Books: newBooks, isLoading: false });
+          setData({ books: newBooks, isLoading: false });
         } else {
-          setData({ Books, isLoading: false, fetchError: true });
+          setData({ books, isLoading: false, fetchError: true });
         }
       } catch (e) {
         // eslint-disable-next-line no-console
         console.error('Failed to fetch Books:', e);
-        setData({ Books, isLoading: false, fetchError: true });
+        setData({ books, isLoading: false, fetchError: true });
       }
     };
     fetchBooks();
-  }, [id]);
+  });
 
   if (isLoading) {
     return <p>Loading...</p>;
--- a/docs/recipes/Making-An-Api-Call.md
+++ b/docs/recipes/Making-An-Api-Call.md
@@ -102,7 +102,7 @@ The modules reducer would handle those dispatched actions so the module would be
 
 When adding response data to the redux store it can be tempting to try to access this directly in other modules. This is **not** a recommended approach as you should aim to have modules as independent as possible.
 
-You could also choose to bubble up your data fetching the the root module and pass down the data as props, which is a common approach with React applications, however this will result in closer coupling between child and root modules.
+You could also choose to bubble up your data fetching to the root module and pass down the data as props, which is a common approach with React applications, however this will result in closer coupling between child and root modules.
 
 ## Fetchye
 
--- a/docs/recipes/Making-An-Api-Call.md
+++ b/docs/recipes/Making-An-Api-Call.md
@@ -100,7 +100,7 @@ The modules reducer would handle those dispatched actions so the module would be
 
 ## Duplicate Calls Across modules
 
-When adding response data to the redux store it can be tempting to try to \access this directly in other modules. This is **not** a recommended approach as you should aim to have modules as independent as possible.
+When adding response data to the redux store it can be tempting to try to access this directly in other modules. This is **not** a recommended approach as you should aim to have modules as independent as possible.
 
 You could also choose to bubble up your data fetching the the root module and pass down the data as props, which is a common approach with React applications, however this will result in closer coupling between child and root module.
 
@@ -276,4 +276,6 @@ export default BooksModule;
 
 Please note that this low config approach relies on the conventions shown above. If the reducer or provider is not setup correctly in the root module you will not benefit from the caching. The `fetchye-one-app` helpers are designed to meet the majority of use cases and may not meet your requirements. It is possible to have a custom configuration using the `fetchye-redux-provider` and `fetchye-immutable-cache` however this could lead to cache misses and unutilized server side calls by modules not using the same configuration.
 
+<!--ONE-DOCS-HIDE start-->
 [☝️ Return To Top](#Making-An-Api-Call)
+<!--ONE-DOCS-HIDE end-->
--- a/docs/recipes/Making-An-Api-Call.md
+++ b/docs/recipes/Making-An-Api-Call.md
@@ -4,15 +4,276 @@
 
 # Making An API Call
 
-Recipe is forthcoming.
+> **TLDR**: Use [Fetchye](#fetchye) with the `fetchye-one-app` helpers.
 
-## POST
-To enable post set [`ONE_ENABLE_POST_TO_MODULE_ROUTES`](../api/server/Environment-Variables.md#ONE_ENABLE_POST_TO_MODULE_ROUTES) environment variable.
-Request body must be either a JSON object or FormData of less than 15KB in size and is passed as props to your module.
+Making API calls within a One App module has some additional considerations over a
+traditional client side React application.
 
-Supported media types:
-- `application/json`
-- `application/x-www-form-urlencoded`
+A basic data fetching example in a client side React app might look like the following:
 
+```jsx
+const BooksModule = () => {
+  const [{ Books, isLoading, fetchError }, setData] = useState({});
+
+  useEffect(() => {
+    const fetchBooks = async () => {
+      try {
+        setData({ Books, isLoading: true, fetchError: false });
+        const response = await fetch('https://some-data-server.com/api/v1/books');
+        if (response.ok) {
+          const newBooks = await response.json();
+          setData({ Books: newBooks, isLoading: false });
+        } else {
+          setData({ Books, isLoading: false, fetchError: true });
+        }
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.error('Failed to fetch Books:', e);
+        setData({ Books, isLoading: false, fetchError: true });
+      }
+    };
+    fetchBooks();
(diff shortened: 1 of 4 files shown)


docs/recipes/Making-An-Api-Call.md
```diff
@@ -4,15 +4,276 @@
...
+  loadModuleData,
+  reducer,
+};
+```
+
+The modules reducer would handle those dispatched actions so the module would be able to retrieve the data from the Redux store.
+
+## Duplicate Calls Across modules
+
+When adding response data to the redux store it can be tempting to try to \access this directly in other modules. This is **not** a recommended approach as you should aim to have modules as independent as possible.
```
10xLaCroixDrinker, line 103:
```suggestion
When adding response data to the redux store it can be tempting to try to access this directly in other modules. This is **not** a recommended approach as you should aim to have modules as independent as possible.
```

docs/recipes/Module-Composition.md
```diff
@@ -0,0 +1,63 @@
...
+<!--ONE-DOCS-HIDE end-->
+
+# Module Composition
+
+A key part of working with One App is module composition. Modules can be rendered inside one another by either using `ModuleRoute` or `RenderModule`.
+
+## `ModuleRoute`
+
+`ModuleRoute` from `holocron-module-route` allows modules to dynamically load other modules
+as a child route. This can be done on any module as routing is not limited to the route module:
```
10xLaCroixDrinker, line 12:
```suggestion
as a child route. This can be done on any module as routing is not limited to the root module:
```

docs/recipes/Module-Composition.md
```diff
@@ -0,0 +1,63 @@
...
+
+## `RenderModule`
+
+Holocron's `RenderModule` provides an alternate method to rendering another module:
+
+```jsx
+const ParentModule = () => (
+  <div>
+    <h1>I am the parent module</h1>
+    <RenderModule moduleName="ChildModule" props={{ greeting: 'hello' }} />;
```
10xLaCroixDrinker, line 38: Don't we want to encourage the convention of kebab-case for module names?

10xLaCroixDrinker, line 38: Don't we want to encourage the convention of kebab-case for module names?

docs/recipes/Module-Composition.md
```diff
@@ -0,0 +1,63 @@
...
+  </div>
+);
+```
+
+To use RenderModule we need to ensure that the module bundle has been correctly
```
10xLaCroixDrinker, line 43:
```suggestion
To use ￼`RenderModule` we need to ensure that the module bundle has been correctly
```

docs/recipes/Module-Composition.md
```diff
@@ -0,0 +1,63 @@
...
+loaded into our client or server before it can be rendered. We can do this by
```
10xLaCroixDrinker, line 44: All server bundles are already stored in memory. On the server it's only the module's data we would need to load.

docs/recipes/Module-Composition.md
```diff
@@ -0,0 +1,63 @@
...
+};
+```
+
+Both `loadModule` and `composeModules` will ensure that the modules client or server bundle has been loaded and the module can be rendered, however there are some minor differences between the two.
+
+`loadModule` takes the modules name as a single argument and will only ensure that module bundle has been loaded so the module can be successfully rendered.
+
+`composeModules` takes an array of objects, each one contains the name of the module required to load. It will then perform `loadModule` on each of those modules and in addition ensure that every modules `loadModuleData` is invoked. This is very important when server side rendering your modules as `loadModuleData` executes the asynchronous tasks that might be required to correctly render a module on the server.
+
+[☝️ Return To Top](#Making-An-Api-Call)
```
10xLaCroixDrinker, line 63: Do we need the one-docs-hide comment on these 'return to top' links?

docs/recipes/Post-To-Modules.md
```diff
@@ -0,0 +1,15 @@
+<!--ONE-DOCS-HIDE start-->
+[👈 Return to Overview](./README.md)
+<!--ONE-DOCS-HIDE end-->
+
+# POST To Module Routes
```
10xLaCroixDrinker, line 5: Is this being fleshed out in a follow up story?

10xLaCroixDrinker (changes requested):

docs/recipes/Post-To-Modules.md
JAdshead: This was just moved out as It does not make sense to be part of `Making an Api Call`. Can mark as WIP

docs/recipes/Module-Composition.md
JAdshead: updated

docs/recipes/Module-Composition.md
JAdshead: true true

docs/recipes/Module-Composition.md
JAdshead: sure i can add. its not something done in other articles however.

docs/recipes/Post-To-Modules.md
JAdshead: currently marked as a WIP

infoxicator: We said that this section should be moved to the API docs IIRC

docs/recipes/Making-An-Api-Call.md
```diff
@@ -4,15 +4,278 @@
...
+};
+```
+
+The modules reducer would handle those dispatched actions so the module would be able to retrieve the data from the Redux store.
+
+## Duplicate Calls Across modules
+
+When adding response data to the redux store it can be tempting to try to access this directly in other modules. This is **not** a recommended approach as you should aim to have modules as independent as possible.
+
+You could also choose to bubble up your data fetching the the root module and pass down the data as props, which is a common approach with React applications, however this will result in closer coupling between child and root module.
```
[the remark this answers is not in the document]
JAdshead: maybe children and root module. there can only be one root module.

infoxicator (approved):

docs/recipes/Making-An-Api-Call.md
```diff
@@ -4,15 +4,278 @@
...
+  if (fetchError) {
+    return <p>Error fetching Books</p>;
+  }
+
+  return (
+    <div>
+      <h1>Books</h1>
+      <ul>
+        {
+          books && books.map((book) => (
```
Francois-Esquire, line 51: Is this supposed to be `Books`?

docs/recipes/Making-An-Api-Call.md
```diff
@@ -4,15 +4,278 @@
...
+          setData({ Books, isLoading: false, fetchError: true });
+        }
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.error('Failed to fetch Books:', e);
+        setData({ Books, isLoading: false, fetchError: true });
+      }
+    };
+    fetchBooks();
+  }, [id]);
```
Francois-Esquire, line 36: Where is `id` coming from?

docs/recipes/Module-Composition.md
```diff
@@ -0,0 +1,65 @@
...
+
+## `RenderModule`
+
+Holocron's `RenderModule` provides an alternate method to rendering another module:
+
+```jsx
+const ParentModule = () => (
+  <div>
+    <h1>I am the parent module</h1>
+    <RenderModule moduleName="child-module" props={{ greeting: 'hello' }} />;
```
Francois-Esquire, line 38:
```suggestion
    <RenderModule moduleName="child-module" props={{ greeting: 'hello' }} />
```
Was the semicolon intentional?

docs/recipes/Making-An-Api-Call.md
```diff
@@ -4,15 +4,278 @@
...
+};
+```
+
+The modules reducer would handle those dispatched actions so the module would be able to retrieve the data from the Redux store.
+
+## Duplicate Calls Across modules
+
+When adding response data to the redux store it can be tempting to try to access this directly in other modules. This is **not** a recommended approach as you should aim to have modules as independent as possible.
+
+You could also choose to bubble up your data fetching the the root module and pass down the data as props, which is a common approach with React applications, however this will result in closer coupling between child and root modules.
```
Francois-Esquire, line 105:
```suggestion
You could also choose to bubble up your data fetching to the root module and pass down the data as props, which is a common approach with React applications, however this will result in closer coupling between child and root modules.
```

docs/recipes/Making-An-Api-Call.md
JAdshead: ah missed that

docs/recipes/Making-An-Api-Call.md
JAdshead: or the others should be `books`

Francois-Esquire (approved):

JamesSingleton (approved):

docs/recipes/Making-An-Api-Call.md
```diff
@@ -4,15 +4,278 @@
...
-## POST
-To enable post set [`ONE_ENABLE_POST_TO_MODULE_ROUTES`](../api/server/Environment-Variables.md#ONE_ENABLE_POST_TO_MODULE_ROUTES) environment variable.
-Request body must be either a JSON object or FormData of less than 15KB in size and is passed as props to your module.
+Making API calls within a One App module has some additional considerations over a
+traditional client side React application.
 
-Supported media types:
-- `application/json`
-- `application/x-www-form-urlencoded`
+A basic data fetching example in a client side React app might look like the following:
```
nellyk, line 12:
```suggestion
A basic data fetching example in a client side React app using the [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) might look like the following:
```

docs/recipes/Making-An-Api-Call.md
```diff
@@ -4,15 +4,278 @@
...
+    } else {
+      store.dispatch({ type: 'FAILED_API' });
+    }
+  } catch (e) {
+    store.dispatch({ type: 'FAILED_API' });
+  }
+};
+
+Books.holocron = {
+  // Runs on both Server and Browser
```
nellyk, line 93:
```suggestion
  // Runs on both Server and Browser. Use global.BROWSER to alternate between the two
```

docs/recipes/Making-An-Api-Call.md
```diff
@@ -4,15 +4,278 @@
...
+  loadModuleData,
+  reducer,
+};
+```
+
+The modules reducer would handle those dispatched actions so the module would be able to retrieve the data from the Redux store.
```
nellyk, line 99:
```suggestion
The modules reducer would handle those dispatched actions so the module would be able to retrieve the data from the Redux store. Depending on your case this value is not mandatory
```

docs/recipes/Making-An-Api-Call.md
```diff
@@ -4,15 +4,278 @@
...
+Updating the first example to use `useFetchye` reduces the amount of boilerplate required for handling the request, loading and error states.
+
+```jsx
+import { useFetchye } from 'fetchye';
+
+const BooksModule = () => {
+  const { isLoading, data, error } = useFetchye('https://some-data-server.com/api/v1/books');
+  const books = data && data.body;
+
+  if (isLoading) {
```
nellyk, line 136: should we also advice using `data.body.status===200`

docs/recipes/Making-An-Api-Call.md
JAdshead: i think in this context that could be confusing

docs/recipes/Making-An-Api-Call.md
JAdshead: to fetch this on the server and have the data propagate to the ui using the reducer would be required

docs/recipes/Making-An-Api-Call.md
JAdshead: we could add a check for `data.ok` but that is something i expect an engineer to handle, this is just a basic example of the fetchye api

10xLaCroixDrinker (approved):

JamesSingleton (approved):

PR #405, opened 2021-01-12
docs(lifecycle): server startup and configuration

Francois-Esquire:
## Description
Add docs on life cycle and configuration. The guide discusses what is happening during each phase of One App starting up and goes over the steps taken to server side render.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,157 @@
...
+When the One App `node.js` server first boots up, there are a series of steps
+undertaken. The first step One App takes is configuration, which can be done via
+command line arguments and environment variables.
+
+**Configuration**
+
+Whenever One App is ran `NODE_ENV` must be defined with either
+`development` or `production`. One App behaves differently depending
+on which node environment it is running under. In development mode,
+One App spins up a development CDN and API server that run on
```
JamesSingleton, line 28: We should link to https://github.com/americanexpress/one-app-dev-cdn here

JamesSingleton (changes requested):

docs/overview/Lifecycle.md
infoxicator, line 28:
```suggestion
One App spins up a development CDN and API server that runs on
```

docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,157 @@
...
+>  * [`ONE_CLIENT_CSP_REPORTING_URL`](../api/server/environment-variables#one_client_csp_reporting_url) ⚠️
+
+### Holocron Runtime
+
+Before One App can start rendering HTML documents from the server, the app will
+preload every Holocron module defined in the Holocron module map. This module map
+is comprised of every module to be used by One App, including the root module which
+serves as our entry point.
+
+```json
```
infoxicator, line 71: maybe link to the module map schema rather than an incomplete example?

Francois-Esquire: I've updated (10a7806)

docs/overview/Lifecycle.md
Francois-Esquire: I removed this portion and discussed it under `development environment` (10a7806)

Francois-Esquire: Updated with links (10a7806)

docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,335 @@
...
+>  * [`HTTPS_PORT`](../api/server/environment-variables#https_port) ⚠️
+>  * [`HTTPS_PRIVATE_KEY_PASS_FILE_PATH`](../api/server/environment-variables#https_private_key_pass_file_path) ⚠️
+>  * [`HTTPS_PRIVATE_KEY_PATH`](../api/server/environment-variables#https_private_key_path) ⚠️
+>  * [`HTTPS_PUBLIC_CERT_CHAIN_PATH`](../api/server/environment-variables#https_public_cert_chain_path) ⚠️
+>  * [`HTTPS_TRUSTED_CA_PATH`](../api/server/environment-variables#https_trusted_ca_path) ⚠️
+
+### Holocron Runtime
+
+Before One App can start rendering HTML documents from the server, the app will
+preload every Holocron module defined in the Holocron module map. This module map
```
nellyk, line 116: can we also add that the modules locally can be defined under package.json?

docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,335 @@
...
+needs to be configured with a valid CSP.
+
+## Server Side Rendering
+
+When One App has loaded all of our modules and started the `express` server,
+the app becomes ready to render our Holocron modules. There are a few ways that
+One App can render our Holocron modules:
+
+- default render path, using the request URL to render the modules that match
+- partial renders, renders only the selected Holocron module in isolation
```
infoxicator, line 182: partials are more than just modules in isolation, it is a different react render method https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup maybe we should add this note in the expanded explanation of partials below

docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,335 @@
...
+MyModule.childRoutes = () => [
+  <ModuleRoute moduleName="my-module">
+    <ModuleRoute moduleName="my-layout-module">
+      <ModuleRoute moduleName="landing-page-module" path="/" />
+    </ModuleRoute>
+  </ModuleRoute>,
+];
+```
+
+**Data Loading**
```
infoxicator, line 221: What do you think a small mention of fetchye and a link to the recipe?

docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,335 @@
...
+
+One App can be configured to statically render Holocron modules as an independent
+entry point when rendered as a partial, regardless of URL path structures
+defined with `Module.childRoutes` and `ModuleRoute` - this is made possible by
+the micro-frontend driven architecture baked into One App, and can be utilized
+to render modules individually. Partial renders gives us the ability to generate
+emails, templates or any static HTML rendered by a Holocron module.
+
+Both the standard and partial rendering follow the same lifecycle,
+with the key difference being the designated entry module being rendered.
```
infoxicator, line 321: Maybe we should adjust this paragraph, the only difference in partials is dispatching the action to change the rendering method `setPartialOnly` that will set the render method to `https://reactjs.org/docs/react-dom-server.html#rendertostaticmarkup` the rest is the same and the difference mentioned here is managed by the user i.e. "rendering in isolation" is not a feature of One App as such but an implementation detail by the module developers

docs/overview/Lifecycle.md
Francois-Esquire: I think we cover that in the "Getting Started" guide, which would come before this guide (under [`Running One App - Local Configuration`](https://github.com/americanexpress/one-app/blob/main/docs/overview/Getting-Started.md#running-one-app)). Do you think it's worth mentioning again?

docs/overview/Lifecycle.md
Francois-Esquire: Valid points, I will correct this.

docs/overview/Lifecycle.md
Francois-Esquire: The lifecycle docs was intended to focus on One App life cycles (server, holocron, render) and the configuration that drives it - Would mentioning `fetchye` fit better in the `Getting Started` guide? Or do you think that would benefit the topic and keep it focused?

docs/overview/Lifecycle.md
Francois-Esquire: Agree with what you are saying, will update.

docs/overview/Lifecycle.md
Francois-Esquire: I have updated this (401b543)

docs/overview/Lifecycle.md
Francois-Esquire: I have updated (04c4301) and removed the implementation specific "rendering isolation" from describing partial rendering.

docs/overview/Lifecycle.md
nellyk: Aha, thank you that looks good :-)

docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,383 @@
...
+>  * [`HTTP_METRICS_PORT`](../api/server/environment-variables#http_metrics_port)
+
+**Production Environment**
+
+There are mandatory environment variables that need to be supplied to run One App
+and there are others to configure aspects of the One App runtime such as reporting
+URLs for errors and CSP violations that occurred on the client. You will need to
+include the environment variables below to run One App in production mode:
+
+>  * [`ONE_CLIENT_CDN_URL`](../api/server/environment-variables#one_client_cdn_url) ⚠️
```
JAdshead, line 97: these still need to be set in development but they just fallback to a default. i think we clarify that that happens in development.

docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,383 @@
...
+>  * [`HTTP_ONE_APP_DEV_PROXY_SERVER_PORT`](../api/server/environment-variables#http_one_app_dev_proxy_server_port)
+
+There is a metrics server that runs in development and production,
+which tracks certain activities happening internally and reports the
+metrics on demand (during local development `http://localhost:3005/metrics`).
+The metrics server runs on port `3005` by default, to change the port:
+
+>  * [`HTTP_METRICS_PORT`](../api/server/environment-variables#http_metrics_port)
+
+**Production Environment**
```
JAdshead, line 90: I would have production before development as everything that applied to production also mostly applies to development.

docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,383 @@
...
+  * [Text Rendering](#text-rendering)
+
+## One App Server
+
+When the One App server first boots up, there are a series of steps
+undertaken before the server becomes operational - configuring One App
+based on environment settings is the first step, followed by pre-loading
+all the Holocron modules onto the server.
+
+### Server Settings
```
JAdshead, line 43: This very useful but is more to do with configuring the one app server than understanding the lifecycle. This could probably be moved elsewhere and referenced here

docs/overview/Lifecycle.md
```diff
@@ -0,0 +1,383 @@
...
+
+>  * [`HOLOCRON_SERVER_MAX_MODULES_RETRY`](../api/server/environment-variables#holocron_server_max_modules_retry)
+>  * [`HOLOCRON_SERVER_MAX_SIM_MODULES_FETCH`](../api/server/environment-variables#holocron_server_max_sim_modules_fetch)
+>  * [`ONE_MAP_POLLING_MAX`](../api/server/environment-variables#one_map_polling_max)
+>  * [`ONE_MAP_POLLING_MIN`](../api/server/environment-variables#one_map_polling_min)
+
+One App will first fetch the Holocron module map using the provided URL
+and proceeds to load every module defined in the module map into memory.
+If a module failed to load for any reason or when validating, One App will
+log the error and recover gracefully, adding the module to a `block` list until
```
JAdshead, line 143: When this happens during the "start" phase one app will not recover and the server will fail to start. Also not all conditions result in the module being added to the block list.

docs/overview/Lifecycle.md
JAdshead: agreed with @Francois-Esquire lets keep this focused on the lifecycle / flow

JAdshead (changes requested):

docs/overview/Lifecycle.md
Francois-Esquire: Good catch, will update.

docs/overview/Lifecycle.md
Francois-Esquire: Agree, will do.

docs/overview/Lifecycle.md
Francois-Esquire: That works too, I'll update.

docs/overview/Lifecycle.md
Francois-Esquire: Updated in (105df92)

docs/overview/Lifecycle.md
Francois-Esquire: Updated in (7e618c0)

docs/overview/Lifecycle.md
Francois-Esquire: This was a good call IMO, I split the guide into lifecycle and configuration (7e618c0)

docs/overview/Lifecycle.md
Francois-Esquire: Updated (7e618c0)

JamesSingleton (approved):

infoxicator (approved):

nellyk (approved):

JAdshead (approved):

PR #406, opened 2021-01-12
feat(metrics): track intl cache size

JAdshead:
## Description

Track intl cache size

## Motivation and Context

This will allow us to track the changes in lang pack caching overtime

## How Has This Been Tested?

locally using the prod-sample and unit tests

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


__tests__/server/metrics/index.spec.js
```diff
@@ -53,4 +55,6 @@ describe('index', () => {
   it('exports appVersion', () => expect(index).toHaveProperty('appVersion', appVersion));
 
   it('exports holocron', () => expect(index).toHaveProperty('holocron', holocron));
+
+  it('exports intlCacje', () => expect(index).toHaveProperty('intlCache', intlCache));
```
infoxicator, line 59:
```suggestion
  it('exports intlCache', () => expect(index).toHaveProperty('intlCache', intlCache));
```

infoxicator:
@JAdshead I think you mentioned before but can you write here how is the size collected overtime and the workaround so we don't have to use the module map poll

src/server/metricsServer.js
```diff
@@ -20,8 +20,10 @@ import { register as metricsRegister, collectDefaultMetrics } from 'prom-client'
 
 import logging from './utils/logging/serverMiddleware';
 import healthCheck from './middleware/healthCheck';
+import { cacheSizeCollector } from './metrics/intl-cache';
 
 collectDefaultMetrics();
+metricsRegister.registerCollector(cacheSizeCollector);
```
JAdshead, line 26: @infoxicator the register ensures that the cacheSizeCollector is called in the same manor default metrics are collected

infoxicator (approved):

Francois-Esquire (approved):

JamesSingleton (approved):

PR #416, opened 2021-01-26
Fixenv

funtimeerrors:
## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


funtimeerrors:
Need to pause this and fix commit history...

JAdshead:
There is existing validation for the env vars in [runtime.js](https://github.com/americanexpress/one-app/blob/main/src/server/config/env/runTime.js) which uses [env-config-utils](https://github.com/americanexpress/env-config-utils/blob/main/src/preprocessEnvVar.js#L30) to validate, we should either be updating this current validation to batch the errors or replace the functionality so we don’t end up with redundant code.

if you make the changes in runtime rather than create a new file the specs should tell you if you got everything. to delete runtime JS all the tests used for that still need to pass

where the errors are currently raised in https://github.com/americanexpress/one-app/blob/main/src/server/config/env/runTime.js#L199

our goal is to capture and report these validation errors as a single error, something similar to this in runtime.js might work

```js
const preprocessErrors = runTime.reduce((errors, envVar) => {
  try{
    preprocessEnvVar(envVar)
    return errors
  } catch (e) {
    return [...errors, e]
  }
}, []);
if (preprocessErrors.length > 0) {
  // throw as a single error...
}
```

PR #418, opened 2021-01-27
Gathered the error checks from argv.js, runTime.js into precheck.js and to be reported in a batch

funtimeerrors:
## Description

precheck.js now more or less does the same validation that argv and runTime do, but instead of throwing one at a time, it collects found errors into an AggregateErrror object and throws it if and when all potential startup errors have been found.  
Had to add npm package es-aggregate-error to shim, because AggregateError isn't supported below Node 15.
## Motivation and Context

If there are multiple errors with One App startup environment, user would only be notified of them one at a time, by building, fixing, and then building again for each error. Now user can (hopefully) fix all startup errors at once.

https://jira.aexp.com/jira/browse/ONE-7287
## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


JAdshead:
This approach duplicated the logic in runtime.js rather than complementing it. Alternative solutions were posted on a PR from this branch previously. 

https://github.com/americanexpress/one-app/pull/416#issuecomment-767715765

PR #433, opened 2021-02-18
fix(dev-holocron-cdn): allow no local module map

JAdshead:
## Description

When running locally it was assumed that there would be locally served modules. If there were not then one-app would fail to start. The one-app server should be able to run without locally served modules

## Motivation and Context

The one-app server should be able to run without locally served modules

## How Has This Been Tested?

Unit tests updated and tested by starting the app with locally served modules and without.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


src/server/devHolocronCDN.js
```diff
@@ -17,19 +17,21 @@
...
+const hasLocalModuleMap = () => fs.existsSync(path.join(process.cwd(), 'static', 'module-map.json'));
+
 const app = express();
 
 app.use(cors());
 app.use('/static', oneAppDevCdn({
   localDevPublicPath: path.join(__dirname, '../../static'),
   remoteModuleMapUrl: argv.moduleMapUrl,
-  useLocalModules: true,
+  useLocalModules: hasLocalModuleMap(),
```
Francois-Esquire, line 34: Should we have a check to see if both `moduleMapUrl` and `localModuleMap` are missing? Or will `moduleMapUrl` always be defined (`http://localhost:3001` IIRC)?

JamesSingleton: I believe it is always defined as you don't have to provide that normally.

JAdshead: moduleMapUrl is required if no locally served modules. this check is done in both oneAppDevCdn and argv.js
https://github.com/americanexpress/one-app/blob/main/src/server/config/env/argv.js#L60

infoxicator (approved):

nellyk (approved):

10xLaCroixDrinker (approved):

Francois-Esquire (approved):

PR #446, opened 2021-03-08
fix(pwa): cache only responses with a 200 status

JamesSingleton:
## Description

Cache only responses with a status of 200.

## Motivation and Context

One App should not cache responses that are not a 200.

## How Has This Been Tested?

Unit test

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/src/client/service-worker/events/utility.js
+++ b/src/client/service-worker/events/utility.js
@@ -127,7 +127,12 @@ export function fetchCacheResource(event, meta) {
   return match(event.request.clone(), { cacheName: meta.cacheName })
     .then(
       (cachedResponse) => cachedResponse
-      || fetch(event.request.clone()).then(setCacheResource(event, meta))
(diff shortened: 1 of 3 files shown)


src/client/service-worker/events/utility.js
```diff
@@ -127,7 +127,12 @@ export function fetchCacheResource(event, meta) {
   return match(event.request.clone(), { cacheName: meta.cacheName })
     .then(
       (cachedResponse) => cachedResponse
-      || fetch(event.request.clone()).then(setCacheResource(event, meta))
+      || fetch(event.request.clone()).then((response) => {
+        if (response.status === 200) {
```
infoxicator, line 131: Wondering if we have to worry about other success status codes like what happens with 304 for example?🤔

JamesSingleton: It would be cached initially right? 🤔

Issue #36, opened 2020-02-13
[QUESTION] 404: Not found in the terminal


giulianok:
I cloned **One App** and created an empty One Module from the Yeoman Generator. The `localhost:3000` is rendering "Welcome to One App", looks like it's working fine, but I'm getting this error in the terminal, not sure why:

```
➜  one-app git:(master) npm start -- --root-module-name=one-module

> @americanexpress/one-app@5.0.0-rc1 start /Users/giulianokranevitter/Projects/Tests/one-app
> node lib/server/index.js "--root-module-name=one-module"

log: one-app-dev-cdn only using locally served modules
✔ 👕 one-app-dev-cdn server listening on port 3001
✔ 👖 one-app-dev-proxy server listening on port 3002
✔ 📊 Metrics server listening on port 3005
✔ 🌎 One App server listening on port 3000
✔ Polled Holocron Module map: no updates (polling again in 7s)
✔ Polled Holocron Module map: no updates (polling again in 9s)
✔ Polled Holocron Module map: no updates (polling again in 11s)
warn: CSP Violation: undefined:undefined:undefined on page http://localhost:3000/ violated the font-src policy via data
error: [
  {
    msg: '404: Not found',
    stack: 'Error: 404: Not found\n' +
      '    at Object.onEnter (/Users/giulianokranevitter/Projects/Tests/one-app/lib/universal/routes.js:49:62)\n' +
      '    at Array.transitionHook (/Users/giulianokranevitter/Projects/Tests/one-app/node_modules/@americanexpress/one-app-router/lib/TransitionUtils.js:48:12)\n' +
      '    at /Users/giulianokranevitter/Projects/Tests/one-app/node_modules/@americanexpress/one-app-router/lib/TransitionUtils.js:119:19\n' +
      '    at /Users/giulianokranevitter/Projects/Tests/one-app/node_modules/@americanexpress/one-app-router/lib/TransitionUtils.js:89:7\n' +
      '    at next (/Users/giulianokranevitter/Projects/Tests/one-app/node_modules/@americanexpress/one-app-router/lib/AsyncUtils.js:39:12)\n' +
      '    at loopAsync (/Users/giulianokranevitter/Projects/Tests/one-app/node_modules/@americanexpress/one-app-router/lib/AsyncUtils.js:56:3)\n' +
[...]

PR #508, opened 2021-05-10
fix(clientErrorLogger): add metaData to server side reported error

infoxicator:
This needs to be merged first 
https://github.com/americanexpress/one-app-ducks/pull/32

Reverts https://github.com/americanexpress/one-app/pull/382
fixes https://github.com/americanexpress/one-app/issues/36#issuecomment-587152336

Adding `util.inspect` to `one-app-ducks` and the `clientErrorReporter` bypasses the production formatter https://github.com/americanexpress/one-app/blob/main/src/server/utils/logging/production-formatter.js 

This has caused an issue trying to serialise the error in monitoring tools.

## Description

Incorrect fixes for this issue caused issues with the production formatter, when the initial report was regarding the development formatter https://github.com/americanexpress/one-app/issues/36#issuecomment-587152336 

This fix ensures the client and server reported errors follow the same schema.

Here is an example of the new server production formatted error:

```
{"schemaVersion":"0.3.0","application":{"name":"One App","version":"5.11.4-2027518b"},"device":{"id":"xx","agent":"Darwin x64"},"level":"error","timestamp":"2021-05-13T17:06:28.386Z","error":{"name":"ServerSideReportedError","message":"404: Not found","stacktrace":"Error: 404: Not found\n    at Object.onEnter"},"metaData":{"location":{"pathname":"/s","search":"","hash":"","action":"POP","key":"h802w0","query":{}},"code":404,"collectionMethod":"applicationError"}}
```

This fix also solves the issue when logging deeply nested objects in development using the development formatter

## How Has This Been Tested?

unit tests

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

No impact

### DIFF
--- a/__tests__/server/middleware/clientErrorLogger.spec.js
+++ b/__tests__/server/middleware/clientErrorLogger.spec.js
@@ -15,13 +15,11 @@
  */
 
 import { createRequest, createResponse } from 'node-mocks-http';
-import util from 'util';
 
 describe('clientErrorLogger', () => {
   let clientErrorLogger;
   jest.spyOn(console, 'error').mockImplementation(() => {});
   jest.spyOn(console, 'warn').mockImplementation(() => {});
-  jest.spyOn(util, 'inspect').mockImplementation(() => {});
 
   function load(nodeEnv) {
     if (typeof nodeEnv !== 'string') {
@@ -80,6 +78,7 @@ describe('clientErrorLogger', () => {
         body: [{
           msg: 'something broke',
           stack: 'Error: something broke\n    at methodA (resource-a.js:1:1)\n    at methodB (resource-b.js:1:1)\n',
+          type: 'ClientReportedError',
           href: 'https://example.com/page-the/error/occurred-on',
           otherData: {
             moduleID: 'dynamic-layout',
@@ -98,10 +97,8 @@ describe('clientErrorLogger', () => {
       console.error.mockClear();
       clientErrorLogger(req, res);
       expect(console.error).toHaveBeenCalledTimes(1);
-      expect(util.inspect).toHaveBeenCalledTimes(1);
-      const logged = util.inspect.mock.calls[0][0];
-      expect(logged).toBeInstanceOf(Object);
-      expect(logged).toHaveProperty('message', 'something broke');
+      const logged = console.error.mock.calls[0][0];
+      expect(logged).toBeInstanceOf(Error);
       expect(logged).toHaveProperty('name', 'ClientReportedError');
       expect(logged).toHaveProperty('stack', 'Error: something broke\n    at methodA (resource-a.js:1:1)\n    at methodB (resource-b.js:1:1)\n');
       expect(logged).toHaveProperty('userAgent', 'Browser/9.0 (Computer; Hardware Software 19_4_0) PackMule/537.36 (UHTML, like Salamander) Browser/99.2.5.0 Browser/753.36');
--- a/__tests__/server/middleware/clientErrorLogger.spec.js
+++ b/__tests__/server/middleware/clientErrorLogger.spec.js
@@ -78,7 +78,6 @@ describe('clientErrorLogger', () => {
         body: [{
           msg: 'something broke',
           stack: 'Error: something broke\n    at methodA (resource-a.js:1:1)\n    at methodB (resource-b.js:1:1)\n',
-          type: 'ClientReportedError',
           href: 'https://example.com/page-the/error/occurred-on',
           otherData: {
             moduleID: 'dynamic-layout',
--- a/__tests__/server/middleware/clientErrorLogger.spec.js
+++ b/__tests__/server/middleware/clientErrorLogger.spec.js
@@ -100,7 +100,8 @@ describe('clientErrorLogger', () => {
       expect(console.error).toHaveBeenCalledTimes(1);
       expect(util.inspect).toHaveBeenCalledTimes(1);
       const logged = util.inspect.mock.calls[0][0];
-      expect(logged).toBeInstanceOf(Error);
+      expect(logged).toBeInstanceOf(Object);
+      expect(logged).toHaveProperty('message', 'something broke');
       expect(logged).toHaveProperty('name', 'ClientReportedError');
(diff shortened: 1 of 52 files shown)


JAdshead (changes requested):

src/server/middleware/clientErrorLogger.js
```diff
@@ -54,8 +54,8 @@ export default function clientErrorLogger(req, res) {
             ...otherData,
             correlationId,
           },
-        });
-        console.error(util.inspect(err, false, 10, true));
+        };
+        console.error(util.inspect(clientError, false, 10, true));
```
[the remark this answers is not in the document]
infoxicator: The issue is not `util.inspect` the same happens if you do a straight `console.log(clientError)` the problem is the Error object shape is somehow is different between node versions

src/server/middleware/clientErrorLogger.js
```diff
@@ -42,11 +41,11 @@ export default function clientErrorLogger(req, res) {
           return;
         }
         const {
-          msg, stack, href, otherData,
+          msg, stack, type, href, otherData,
         } = raw;
         const err = new Error(msg);
         Object.assign(err, {
-          name: 'ClientReportedError',
+          name: type,
```
JAdshead, line 48: as this is the clientErrorLogger, what is the scenario where its not a client reported error ?

infoxicator: I can revert to hardcoding this value, however I need to pass a new value `ServerSideReportedError` here https://github.com/americanexpress/one-app-ducks/pull/32/files#diff-5dc992175b262921c364a57cee1fcd0f6e81cd4d086d8115a20238ca7e14b119R40 so we can adjust the schema accordingly in the production logger here: https://github.com/americanexpress/one-app/pull/508/files#diff-481c987f1307f2ec947cbca5fd31e173890266a92b6148ba3f177e90d8136cc0

__tests__/server/utils/logging/production-formatter.spec.js
```diff
@@ -169,6 +169,34 @@ describe('production-formatter', () => {
       expect(() => JSON.parse(entry)).not.toThrowError();
       expect(JSON.parse(entry)).toMatchSnapshot();
     });
+
+    it('encodes ServerSideReported Error as parseable JSON', () => {
+      function buildServerSideErrorEntry() {
+        const error = new Error('something broke');
+        Object.assign(error, {
+          name: 'ServerSideReportedError',
```
nellyk, line 177: nitpick ~ might be better to rename this to `type`?

infoxicator: @nellyk don't think we can change it, this is part of our internal schema so it can't be changed, I can share details internally 👍

nellyk (dismissed):

__tests__/server/utils/logging/production-formatter.spec.js
PixnBits: FWIW based off of [`Error.prototype.name`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#instance_properties)

src/server/utils/logging/production-formatter.js
```diff
@@ -86,6 +86,10 @@ function leadingErrorWithOtherArgs(level, ...args) {
     };
   }
 
+  if (err.name === 'ServerSideReportedError') {
```
PixnBits, line 89: is this dependent on the instance of the Error or can duck typing(?) be used instead? _maybe_
```suggestion
  if (err.metaData) {
```

src/server/middleware/clientErrorLogger.js
PixnBits: Sorry, I'm not sure I follow 😓 : what's the need behind overriding the Error instance's name property with `ServerSideReportedError`?

PixnBits (changes requested):

src/server/middleware/clientErrorLogger.js
infoxicator: Reason was to identify the difference between the `clientReportedError` and the server error so we could append the metadata, however, your suggestion of doing it by checking for the metaData property, however, for the client reported error the metaData will appear twice.

`if (err.metaData) {`

Inside the `err.request.metaData` and `err.metaData`

src/server/utils/logging/production-formatter.js
infoxicator: This will duplicate the metaData for the clientReported error `err.request.metaData and err.metaData`

__tests__/server/middleware/clientErrorLogger.spec.js
```diff
@@ -80,6 +78,7 @@ describe('clientErrorLogger', () => {
         body: [{
           msg: 'something broke',
           stack: 'Error: something broke\n    at methodA (resource-a.js:1:1)\n    at methodB (resource-b.js:1:1)\n',
+          type: 'ClientReportedError',
```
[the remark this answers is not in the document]
infoxicator: The value was a hardcoded string in the `clientErrorReporter`, not supplied by the ducks

src/server/middleware/clientErrorLogger.js
infoxicator: We don't have to use the `name` property since you mention it could cause confusion and debugging issues with the prototype. What about using another property to identify if it is an error reported by the client or the server? like `errorType`?

__tests__/server/utils/logging/production-formatter.spec.js
```diff
@@ -169,6 +169,33 @@ describe('production-formatter', () => {
       expect(() => JSON.parse(entry)).not.toThrowError();
       expect(JSON.parse(entry)).toMatchSnapshot();
     });
+
+    it('encodes Server Rerported Error as parseable JSON with nested metadata objects', () => {
```
PixnBits, line 173:
```suggestion
    it('encodes Server Reported Error as parseable JSON with nested metadata objects', () => {
```

PixnBits (changes requested):
one nitpick of spelling 😓

PixnBits (dismissed):

PixnBits:
be sure to use an updated commit message (and maybe edit the PR title?) as the original is no longer the change at hand

nellyk (approved):

JAdshead (approved):

PR #534, opened 2021-07-16
fix(helmet): fixes race condition with react-helmet

e-l-i-s-e:
I'm proposing that when One App runs `initClient`  that it only remove helmet scripts from the body and will no longer append those scripts to the head of the html doc. Appending the helmet scripts to the head will be done by `Helmet`, because `Helmet` will add the scripts to the head whether it runs before or after `moveHelmetScripts`.

## Description

This PR changes `moveHelmetScripts` so that 1) it only queries the DOM for helmet scripts in the body rather than in the entire document, 2) no longer appends scripts to the head. 

## Motivation and Context

When One App middleware runs `sendHtml` on the server and creates an html doc it appends helmet scripts from the request to the html body. When the app is initialized on the client, the `moveHelmetScripts` function is invoked and it retrieves all scripts in the html doc, removes them from the body and adds them to the head.

The problem is that by the time this runs, it is possible for the `Helmet` component from `react-helmet` to have appended scripts to the head of the html doc. That duplicates the helmet scripts and will cause `moveHelmetScripts` to throw an error because it will try to remove all scripts it finds from the body without checking if they were found in the body. 

There are 2 scenarios I've found where Helmet appends scripts before `moveHelmetScripts` runs: 
1) When many modules are loaded on the page (race condition), `moveHelmetScripts` is firing after `Helmet` appends the scripts to the head. This seems to be due to the fact that it's invoked in an event listener waiting for `DOMContentLoaded` that can sometimes run after`react-helmet` updates the DOM (which is invoked in a componentWillMount lifecycle function). After `Helmet` appends scripts to the head, then `moveHelmetScripts` runs and grabs scripts from the body and the head. It will attempt to remove the head scripts from the body and then throws the error: `Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.`

2) The second scenario when `moveHelmetScripts` throws an error is when a dev wants `react-helmet` to update the DOM ASAP and passes `defer={false}` to `Helmet`. https://github.com/nfl/react-helmet#readme
If defer is false, then `Helmet` appends scripts to the head much earlier and will always cause `moveHelmetScript` to error if scripts have been added to the body on the server. Here is an example with the sample module, `ssr-frank` adding one helmet script with `react-helmet`: 
```jsx
<Helmet
  defer={false}
  onChangeClientState={(newState, addedTags, removedTags) => console.log('react-helmet - onChangeClientState', newState, 'added', addedTags, 'removed', removedTags)}
  script={[
      {
        type: 'application/ld+json',
        innerHTML: JSON.stringify({ '@context': 'https://schema.org' }),
       }
  ]}
/>
```
You can see that the script is duplicated and the function throws an error: 
Screen Shot 2021-07-16 at 3 38 17 PM

## How Has This Been Tested?

I ran this using a `demo` route with one module and on a working route with multiple modules without passing the `defer` prop (so it's `true`) and passed it with the value `false`.

If we run `ssr-frank` with the following `Helmet` code added and the `append` command removed, you see that everything looks fine. `moveHelmetScripts` will remove the one script off the html `body` and then `react-helmet` will add the one script to the head: 
Screen Shot 2021-07-16 at 3 40 39 PM

If I pass `defer={false}` to `Helmet`, then the script will be added to the head before `moveHelmetScripts` runs. Everything works: 
Screen Shot 2021-07-16 at 3 32 24 PM

This shouldn't affect any other areas of the code.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

This should have not noticeable impact on current users who do not see the DOM exception error. It will resolve that error for users who are currently seeing it and it will allow developers using One App to instruct `Helmet` to update the DOM earlier.

### DIFF
--- a/src/client/prerender.js
+++ b/src/client/prerender.js
@@ -46,9 +46,16 @@ export function loadPrerenderScripts(initialState) {
 
 export function moveHelmetScripts() {
   document.addEventListener('DOMContentLoaded', () => {
-    const helmetScripts = [...document.querySelectorAll('script[data-react-helmet]')];
-    helmetScripts.forEach((script) => document.body.removeChild(script));
-    helmetScripts.forEach((script) => document.head.appendChild(script));
+    const headHelmetScripts = [...document.head.querySelectorAll('script[data-react-helmet]')];
+    const bodyHelmetScripts = [...document.body.querySelectorAll('script[data-react-helmet]')];
+
+    bodyHelmetScripts.forEach((script) => document.body.removeChild(script));
+
+    // If the helmet scripts exist in the head then we can assume that the body
+    // scripts were already added by react-helmet
+    if (headHelmetScripts.length === 0) {
+      bodyHelmetScripts.forEach((script) => document.head.appendChild(script));
+    }
   });
 }
 
--- a/__tests__/client/__snapshots__/prerender.spec.js.snap
+++ b/__tests__/client/__snapshots__/prerender.spec.js.snap
@@ -1,22 +1,5 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`moveHelmetScripts should only move scripts that were injected by react-helmet, ignoring others: helmet scripts 1`] = `
-NodeList [
-  <script
-    data-react-helmet="true"
-    src="hello.js"
-  />,
-  <script
-    data-react-helmet="true"
-    src="world.js"
-  />,
-  <script
-    data-react-helmet="true"
-    src="baz.js"
-  />,
-]
-`;
-
 exports[`moveHelmetScripts should only move scripts that were injected by react-helmet, ignoring others: non-helmet scripts 1`] = `
 NodeList [
   <script
@@ -24,6 +7,15 @@ NodeList [
   />,
   <script
     src="bar.js"
+  />,
+]
+`;
+
+exports[`moveHelmetScripts should remove helmet scripts from the body and only append helmet scripts to the head if react-helmet has not already injected them: head helmet scripts 1`] = `
+Array [
+  <script
+    data-react-helmet="true"
+    src="hello.js"
   />,
 ]
 `;
--- a/__tests__/client/prerender.spec.js
+++ b/__tests__/client/prerender.spec.js
@@ -167,7 +167,21 @@ describe('moveHelmetScripts', () => {
     document.addEventListener.mock.calls[0][1]();
 
     expect(document.querySelectorAll('body script')).toMatchSnapshot('non-helmet scripts');
-    expect(document.querySelectorAll('head script')).toMatchSnapshot('helmet scripts');
+  });
+
+  it('should remove helmet scripts from the body and only append helmet scripts to the head if react-helmet has not already injected them', () => {
+    const scriptForBody = createScript({ src: 'hello.js', helmet: true });
+    const scriptForHead = createScript({ src: 'hello.js', helmet: true });
+
+    document.body.appendChild(scriptForBody);
+    document.head.appendChild(scriptForHead);
+
+    moveHelmetScripts();
+    document.addEventListener.mock.calls[0][1]();
+    const scriptsFromHead = [...document.head.querySelectorAll('script')];
+    expect([...document.body.querySelectorAll('script')].length).toEqual(0);
+    expect(scriptsFromHead).toMatchSnapshot('head helmet scripts');
+    expect(scriptsFromHead.length).toEqual(1);
   });
 
   it('should not expect NodeList to have forEach due to IE', () => {


JAdshead:
PR Labeler currently has an issue with forks, I will add manually - https://github.com/TimonVS/pr-labeler-action/issues/25

src/client/prerender.js
```diff
@@ -46,9 +46,8 @@ export function loadPrerenderScripts(initialState) {
 
 export function moveHelmetScripts() {
   document.addEventListener('DOMContentLoaded', () => {
-    const helmetScripts = [...document.querySelectorAll('script[data-react-helmet]')];
+    const helmetScripts = [...document.body.querySelectorAll('script[data-react-helmet]')];
     helmetScripts.forEach((script) => document.body.removeChild(script));
-    helmetScripts.forEach((script) => document.head.appendChild(script));
```
JAdshead, line 51: so think the one risk with this is that there could be small amount time where the scripts are removed and Helmet has not added them back. what are your thought on the following solution - https://github.com/americanexpress/one-app/commit/cdefce96227e7558931c203a153dbc540c33a230
By delaying render until after the scripts have been moved it should negate the race condition.

10xLaCroixDrinker:
@e-l-i-s-e thank you for your contribution! after looking at the issue, we were thinking something more like this: https://github.com/americanexpress/one-app/commit/c0e17b19eed6d2bfb9135e5c579a955d80a9fb74

This way we can make sure that the scripts are moved immediately instead of waiting for react-helmet to do it

Is that a change you'd be amenable to make?

10xLaCroixDrinker (changes requested):
@e-l-i-s-e thank you for your contribution! after looking at the issue, we were thinking something more like this: c0e17b1

This way we can make sure that the scripts are moved immediately instead of waiting for react-helmet to do it

Is that a change you'd be amenable to make?

src/client/prerender.js
e-l-i-s-e: That is an excellent point and I wish that I had thought of it! :)

e-l-i-s-e:
@10xLaCroixDrinker I think that you both raise a very important point. I want to think for a bit more about the solution you have proposed and test it out on a few things if you don't mind. I'll get back to you shortly. Thanks again for providing a possible solution!

e-l-i-s-e:
Hello, @10xLaCroixDrinker and @JAdshead -- I have a few questions that I haven't yet been able to answer myself and wondered if either of you know the answers:
1. The Holocron module in our app that adds Helmet scripts was rendering 4 or 5 times before the `domContentLoaded` event was firing (and invoking `moveHelmetScripts`) with the initial solution I proposed. So `react-dom` must be using a different strategy to determine when it can render/hydrate that allows it to run slightly sooner? Would we want to imitate the way that `react-dom` is doing? Or do you think that the difference is insignificant/not worth the effort required? 
2. Shouldn't delaying `reactDOM.render()` or `reactDOM.hydrate()` increase TTI by some small amount? I assumed that it would, but I wasn't able to see a noticeable difference when I measured the speed of the pages I was testing. Maybe the difference really is just a matter of just a few milliseconds? 

If so, then I think the proposed solution is perfect and will push up changes.

e-l-i-s-e:
I'm thinking that it might be slightly more optimal to do a shallow comparison of scripts queried from the body and head than to delay the app hydrating. I'll push up that change and see what you both think.

JAdshead:
> Hello, @10xLaCroixDrinker and @JAdshead -- I have a few questions that I haven't yet been able to answer yet:
> 
> 1. The Holocron module in our app that adds Helmet scripts was rendering 4 or 5 times before the `domContentLoaded` event was firing (and invoking `moveHelmetScripts`) with the initial solution I proposed. So `react-dom` must be using a different strategy to determine when it can hydrate that allows it to run slightly sooner? Would we want to imitate what`react-dom` is doing? Or do you think that the difference is insignificant/not worth the effort required?
> 2. Shouldn't delaying `reactDOM.render()` or `reactDOM.hydrate()` increase TTI by some small amount? I assumed that it would, but I wasn't able to see a noticeable difference when I measured the speed of the pages I was testing. Maybe the difference really is just a matter of just a few milliseconds?
> 
> If you aren't concerned with that possible impact, then I think the proposed solution is great and I'll push up changes.

I suspect that this is because react render is preventing the domContentLoaded from firing as it updates the dom. if we wait for domContentLoaded to fire before calling render then this should not be an issue. I don't foresee any measurable impact to users for TTI.

e-l-i-s-e:
Hi @JAdshead and @10xLaCroixDrinker, have you had a chance to look at the new code change [afefbc7](https://github.com/americanexpress/one-app/commit/afefbc7f9fa3bc24cffc3cd1e0231a271b9673da)?

src/client/prerender.js
```diff
@@ -46,9 +46,17 @@ export function loadPrerenderScripts(initialState) {
...
+    const headHelmetScripts = [...document.head.querySelectorAll('script[data-react-helmet]')];
+    const bodyHelmetScripts = [...document.body.querySelectorAll('script[data-react-helmet]')];
+
+    bodyHelmetScripts.forEach((script) => document.body.removeChild(script));
+
+    // If the number of helmet scripts in the head equals the number of helmet scripts
+    // the body, then we can assume that those scripts were already added by
+    // react-helmet because one-app would not have added them to the head yet
+    if (headHelmetScripts.length !== bodyHelmetScripts.length) {
```
JAdshead, line 57: I think there are too many potential variables with this approach, its possible that more head helmet scripts are added than originally set in the body in the server render as you can't guarantee that the server bundle and client bundles won't always match up.

e-l-i-s-e: I wonder about that. How would more head helmet scripts get added before this function is called? Wouldn't react-helmet (or equivalent) be the only other library responsible for inserting helmet scripts? Where would the extra helmet scripts come from? One-app wouldn't add a helmet script before this so it couldn't have done that. 

I can't think of a concrete example where that would happen so I would be interested to know what you're thinking of exactly.

e-l-i-s-e:
Hi @JAdshead, It makes sense that more helmet scripts could be added from client bundles. So would it be sufficient to check if any helmet scripts are present in the head in that case? That would mean that another library added them, correct? Since one-app puts the helmet scripts in the body and wouldn't have appended any to the head until this function is evoked. 
Here's the new option: [48b7c42](https://github.com/americanexpress/one-app/commit/48b7c423d366786818622699bf9b2195f61b43ec)

e-l-i-s-e:
Hi @JAdshead and @10xLaCroixDrinker, I thought I'd circle back and see if you had a chance to look my recent comment and commit -- https://github.com/americanexpress/one-app/pull/534#issuecomment-891046983.

Do either or you foresee a problem with that simpler approach? Then we won't need to worry about delaying the rendering/hydrating of the app. If you would prefer to go with your initial proposal, I'm happy to use that instead.

src/client/prerender.js
```diff
@@ -46,9 +46,17 @@ export function loadPrerenderScripts(initialState) {
...
-    helmetScripts.forEach((script) => document.body.removeChild(script));
-    helmetScripts.forEach((script) => document.head.appendChild(script));
+    const headHelmetScripts = [...document.head.querySelectorAll('script[data-react-helmet]')];
+    const bodyHelmetScripts = [...document.body.querySelectorAll('script[data-react-helmet]')];
+
+    bodyHelmetScripts.forEach((script) => document.body.removeChild(script));
+
+    // If the number of helmet scripts in the head equals the number of helmet scripts
+    // the body, then we can assume that those scripts were already added by
+    // react-helmet because one-app would not have added them to the head yet
```
JAdshead, line 56:
```suggestion
    // If the helmet scripts exist in the head then we can assume that the body
    // scripts were already added by react-helmet
```

JAdshead:
I am happy to accept this. Just update the comment to match the flow.

e-l-i-s-e:
Thanks, @JAdshead. I updated the comments and it's ready for re-review.

JAdshead (approved):

marcusrbrown (approved):

10xLaCroixDrinker (approved):

PR #591, opened 2021-10-01
feat(externals): added flag to not throw if externals versions have a mismatch

Matthew-Mallimo:
## Description
Adds a new flag, DANGEROUSLY_ACCEPT_BREAKING_EXTERNALS, that when set `true`, will log a warning instead of throwing an error whenever a mismatch in version for externals is detected.

## Motivation and Context

When upgrading versions of `providedExternals` set by your route module, it can be frustrating because one-app will fail to start up because it throws an error when mismatching external versions are found. If the app is already running, module map polling will fail.

The DANGEROUSLY_ACCEPT_BREAKING_EXTERNALS flag will prevent this from happening by logging a warning instead of throwing an error.

This is a dangerous flag to set because mismatching versions can cause unexpected behaviors for modules, and should only be used during a transition of external versions.

## How Has This Been Tested?

unit tested

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/__tests__/server/utils/onModuleLoad.spec.jsx
+++ b/__tests__/server/utils/onModuleLoad.spec.jsx
@@ -253,6 +253,20 @@ describe('onModuleLoad', () => {
     expect(() => onModuleLoad({ module: { [CONFIGURATION_KEY]: configuration, [META_DATA_KEY]: { version: '1.0.11' } }, moduleName: 'my-awesome-module' })).toThrowErrorMatchingSnapshot();
   });
 
+  it('logs a warning if the root module provides an incompatible version of a required external and DANGEROUSLY_ACCEPT_BREAKING_EXTERNALS is set to true', () => {
+    process.env.DANGEROUSLY_ACCEPT_BREAKING_EXTERNALS = true;
+    RootModule[CONFIGURATION_KEY].providedExternals = {
+      'dep-a': { version: '2.1.0', module: () => 0 },
+    };
+    const configuration = {
+      requiredExternals: {
+        'dep-a': '^2.1.1',
+      },
+    };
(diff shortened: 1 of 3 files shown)


10xLaCroixDrinker:
> ## How Has This Been Tested?
> unit tested

This requires more than unit tests. What scenarios have you run in a development environment?

Matthew-Mallimo:
> > ## How Has This Been Tested?
> > unit tested
> 
> This requires more than unit tests. What scenarios have you run in a development environment?

I have also run this locally. Tested:
* Starting one-app with externals semver mismatch - Flag off - Fails to start
* Starting one-app with externals semver mismatch - Flag on - Starts up with no issues
* Deploying a module with semver mismatch while one-app is running - Flag off - module map polling fails
* Deploying a module with semver mismatch while one-app is running - Flag on - Module map continues to poll successfully while logging the warning to the console.

__tests__/server/utils/onModuleLoad.spec.jsx
```diff
@@ -253,6 +253,20 @@ describe('onModuleLoad', () => {
...
+    };
+    const configuration = {
+      requiredExternals: {
+        'dep-a': '^2.1.1',
+      },
+    };
+    expect(() => onModuleLoad({ module: { [CONFIGURATION_KEY]: configuration, [META_DATA_KEY]: { version: '1.0.10' } }, moduleName: 'my-awesome-module' })).not.toThrow();
+    expect(consoleWarnSpy).toHaveBeenCalledTimes(1);
+  });
+
```
JAdshead, line 269: maybe also include test that checks if `DANGEROUSLY_ACCEPT_BREAKING_EXTERNALS` is unset that it throws the error

Matthew-Mallimo: That is already tested on this line
https://github.com/americanexpress/one-app/blob/main/__tests__/server/utils/onModuleLoad.spec.jsx#L225

PR #611, opened 2021-11-09
feat(runtime): support node 16 with integration tests

10xLaCroixDrinker:
Add support for Node 16

## Description
Update deps and remove incompatible/deprecated fibers

## Motivation and Context
Prep for v6

## How Has This Been Tested?
`npm test`

## Types of Changes

- [x] Dependency update
- [ ] Security update

## Checklist:

- [x] Performance tests should be ran against the server prior to merging.


scripts/build-one-app-docker-setup.js
```diff
@@ -102,7 +104,7 @@ const doWork = async () => {
     );
   }
 
-  const extraCertsExists = await fs.exists(path.join(sampleProdDir, 'extra-certs.pem'));
+  const extraCertsExists = await fs.access(path.join(sampleProdDir, 'extra-certs.pem'));
```
Matthew-Mallimo, line 107: Had to look up why this changed, `fs.exists` is deprecated:
https://nodejs.org/api/fs.html#fsexistspath-callback

Just for future reviewers!

scripts/utils.js
```diff
@@ -39,6 +39,8 @@ const promisifySpawn = (...args) => new Promise((res, rej) => {
   const spawnedProcess = spawn(...args);
   let stdout = '';
   let stderr = '';
+  spawnedProcess.stdout.setEncoding('utf8');
+  spawnedProcess.stderr.setEncoding('utf8');
```
Matthew-Mallimo, line 43: Is this an enhancement or required with the upgrade ?

10xLaCroixDrinker: this is actually left over from some changes to log on `data` instead of at the end. it makes the chunks strings instead of buffers. it can be removed

__tests__/integration/helpers/testRunner.js
```diff
@@ -75,7 +75,7 @@ const setUpTestRunner = async ({
   dockerComposeUpProcess.stderr.pipe(logWatcherDuplex);
 
   // uncomment this line in order to view full logs for debugging
-  // logWatcherDuplex.pipe(process.stdout);
```
infoxicator, line 78: is this meant to be left uncommented or just used when debugging integration tests locally?

10xLaCroixDrinker: It is meant to be uncommon ted while the PR is still a draft

JAdshead:
just the one failing test
```
● heapdump › writes a heapdump to /tmp on a SIGUSR2 signal

    Failed to match: /wrote heapdump out to .+/ in logs

      20 | const timeOutPromise = (timeout, msg) => new Promise((_, rej) => {
      21 |   setTimeout(
    > 22 |     () => rej(new Error(msg)),
         |               ^
      23 |     timeout
      24 |   );
      25 | });
      ```

**Review by Matthew-Mallimo — APPROVED** [review]

File: __tests__/integration/helpers/testRunner.js (hunk 3 above)
Comment (10xLaCroixDrinker) [reply __tests__/integration/helpers/testRunner.js]: Removed

**Comment by 10xLaCroixDrinker** [comment]
> just the one failing test

Resolved 😄

File: src/server/utils/heapdump.js [hunk 4]
```diff
@@ -32,11 +34,12 @@ if (process.execArgv.includes('--report-on-signal')) {
...
+  const heapStream = v8.getHeapSnapshot();
+  const fileStream = fs.createWriteStream(targetFilename);
+
   try {
-    await pipelinePromise(
-      v8.getHeapSnapshot(),
-      fs.createWriteStream(targetFilename)
-    );
+    heapStream.pipe(fileStream);
+    await finishedPromise(fileStream);
```
JAdshead, line 42: No problem with this solution. I don't see anything to indicate that pipeline changed between 12 and 16, out of curiosity whats the reason for this change ?

10xLaCroixDrinker: The promise from `pipeline` wasn't resolving. In node v14 it was updated to only resolve when a `close` event was emitted￼. The option to do that in `createWriteStream` was added in 12 and turned on by default in 14. I did try, as a confidence check, explicitly setting the `emitClose` option to `true`, but it had no effect. I switched to this since we're not doing any transforms so `pipeline` isn't really necessary, and found this implementation didn't have the same issues.

JAdshead (approved):

10xLaCroixDrinker:
I'll open a new one soon that has the rest of the v6 changes

PR #631, opened 2021-12-08
fix(holocron): bad modules could cause crashes and prevent restart

Matthew-Mallimo:
## Description

Updates holocron to v1.2.0, which includes checks to prevent bad modules from stopping module map polling.
Adds integration tests to an upcoming change to holocron

## Motivation and Context

This new version of holocron will prevent one-app from crashing when a bad module is in the module map. 

On one-app start, the server will log an error stating the bad module will be ignored until the next poll where the module succeeds to load.

If one-app is already running, it will log an error stating the bad module's name, and revert back to a working module.

## How Has This Been Tested?

integration tests cover:
* one-app start up with a bad module
	* Should start up normally, just ignoring the bad module
* one-app running with moduleA at 0.0.0, and moduleA 0.0.1 is deployed with problems
	* one-app should continue to poll ignoring v0.0.1 until its healthy
* one-app running when a new module is deployed with issues
	* should continue to poll and ignore the broken module until its healthy

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


__tests__/integration/one-app.spec.js
```diff
@@ -55,6 +57,51 @@ yargs.array('scanEnvironment');
...
+      const [,
+        problemModule,
+        problemModuleUrl,
+        workingUrl,
+      ] = revertErrorMatch.exec(loggedError);
+      const gitSha = await retrieveGitSha();
+      await expect(requiredExternalsError).resolves.toMatch(revertErrorMatch);
+      expect(problemModule).toBe(moduleName);
+      expect(problemModuleUrl).toBe(`${testCdnUrl}/${gitSha}/${moduleName}/${version}/${moduleName}.node.js`);
+      // eslint-disable-next-line no-useless-escape
```
infoxicator, line 94: why is this needed for the next line?

__tests__/integration/one-app.spec.js
```diff
@@ -71,6 +118,7 @@ describe('Tests that require Docker setup', () => {
 
     beforeAll(async () => {
       removeModuleFromModuleMap('late-frank');
+      removeModuleFromModuleMap('unhealthy-frank');
```
infoxicator, line 121: is this required here? 'unhealthy-frank' shouldn't be in the module map because it is a brand new module map inside this describe block?

__tests__/integration/one-app.spec.js
```diff
@@ -700,30 +748,72 @@ describe('Tests that require Docker setup', () => {
...
-            await expect(requiredExternalsError).resolves.toMatch(requiredExternalsErrorMatch);
+          test('fails to get external `react-intl` for child module as an unsupplied `requiredExternal` - Logs reverting message', async () => {
+            const revertErrorMatch = /There was an error loading module (?<moduleName>.*) at (?<url>.*). Reverting back to (?<workingModule>.*)/;
+            const requiredExternalsError = searchForNextLogMatch(revertErrorMatch);
+            await addModuleToModuleMap({
+              moduleName,
+              version,
+            });
+            // not ideal but need to wait for app to poll;
+            await waitFor(5000);
```
infoxicator, line 780: Are we loading the same module `'cultured-frankie';` again? if so we should try and merge this test with the one above if you are able to get both error messages at the same time. This is to avoid overloading the integration tests and make them take longer unnecessarily

prod-sample/sample-modules/unhealthy-frank/0.0.0/README.md
```diff
@@ -0,0 +1,3 @@
+# healthy-frank
```
infoxicator, line 1: unhealthy frank needs a new description

__tests__/integration/one-app.spec.js
```diff
@@ -700,30 +748,72 @@ describe('Tests that require Docker setup', () => {
...
+              workingUrl,
+            ] = revertErrorMatch.exec(loggedError);
+            const gitSha = await retrieveGitSha();
+            await expect(requiredExternalsError).resolves.toMatch(revertErrorMatch);
+            expect(problemModule).toBe('cultured-frankie');
+            expect(problemModuleUrl).toBe(`${testCdnUrl}/${gitSha}/${moduleName}/${version}/${moduleName}.node.js`);
+            // eslint-disable-next-line no-useless-escape
+            expect(workingUrl).toBe(`${testCdnUrl}/${gitSha}/${moduleName}/0.0.0/${moduleName}.node.js\"}`);
+          });
+          test('fails to get external `semver` for child module as an unsupplied `requiredExternal` for new module in mooduleMap', async () => {
```
infoxicator, line 794: have we not tested the same scenario above when requiring `react-intl`? the only difference I can see is `unhealthy-frank` tries to load a different external `semver`?

If this test is to test the scenario when 2 modules with broken required externals are deployed consecutively, then the afterEach in line 754 is resetting the module map anyway and removing `cultured-frankie` from the previous test

prod-sample/sample-modules/unhealthy-frank/0.0.0/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "unhealthy-frank",
```
infoxicator, line 2: I wonder if we can reuse any of the other franks rather than creating a brand new one? If we deploy the correct version of the root module that doesn't provide any externals, I think `cultured-frankie@0.0.1` should be enough?

__tests__/integration/one-app.spec.js
Matthew-Mallimo: I was only able to get one regex matcher to work at a time within a single test case. Thats why I added another test case

__tests__/integration/one-app.spec.js
Matthew-Mallimo: This test case test when a new module is added to the module map with an error. The previous test case tests when a module's version is updated with an issue

prod-sample/sample-modules/unhealthy-frank/0.0.0/package.json
Matthew-Mallimo: It was easier to add a new module that only had one version with an issue, than to manage an existing module with one working version and one broken one

Matthew-Mallimo (commented):
left comments

prod-sample/sample-modules/unhealthy-frank/0.0.0/package.json
infoxicator: I am just thinking that it is more work to maintain another frank but I guess now we removed the package-locks from them we shouldn't need to change them too much.

Is it achievable with the franks that we already got? if so maybe double check with the team to see what's their preference

infoxicator: maintenance aside, building another module would also be expensive and increase integration tests build time

__tests__/integration/one-app.spec.js
infoxicator: What does the log look like? does it log both lines?
`/Failed to get external react-intl from root module/`
and
`/There was an error loading module (?<moduleName>.*) at (?<url>.*). Reverting back to (?<workingModule>.*)`?

If so we should try and find why the regex is not matching in one single case, loading the same module to the module map twice and that 5sec delay is expensive if we can avoid it 👍

__tests__/integration/one-app.spec.js
infoxicator: right so we are testing 1 case when the same module is updated (version bump) and it is broken and also when a brand new module which is broken is added for the first time? is that right?

__tests__/integration/one-app.spec.js
```diff
@@ -700,30 +748,72 @@ describe('Tests that require Docker setup', () => {
...
+            const revertErrorMatch = /There was an error loading module (?<moduleName>.*) at (?<url>.*). Ignoring (?<ignoredModule>.*) until .*/;
+            const requiredExternalsError = searchForNextLogMatch(revertErrorMatch);
+            const modName = 'unhealthy-frank';
+            const modVersion = '0.0.0';
+            await addModuleToModuleMap({
+              moduleName: modName,
+              version: modVersion,
+            });
+            // not ideal but need to wait for app to poll;
+            await waitFor(5000);
```
10xLaCroixDrinker, line 804: Why don't we set a lower max pall time for the integration tests?

__tests__/integration/one-app.spec.js
```diff
@@ -700,30 +748,72 @@ describe('Tests that require Docker setup', () => {
...
+            const [,
+              problemModule,
+              problemModuleUrl,
+              ignoredModule,
+            ] = revertErrorMatch.exec(loggedError);
+            const gitSha = await retrieveGitSha();
+            await expect(requiredExternalsError).resolves.toMatch(revertErrorMatch);
+            expect(problemModule).toBe(modName);
+            expect(problemModuleUrl).toBe(`${testCdnUrl}/${gitSha}/${modName}/${modVersion}/${modName}.node.js`);
+            // eslint-disable-next-line no-useless-escape
```
10xLaCroixDrinker, line 815: Is this needed?

prod-sample/sample-modules/unhealthy-frank/0.0.0/package.json
10xLaCroixDrinker: An additional version of an existing module would have the same downside.

prod-sample/sample-modules/unhealthy-frank/0.0.0/package.json
```diff
@@ -0,0 +1,31 @@
...
+  },
+  "devDependencies": {
+    "@americanexpress/one-app-bundler": "^6.14.1",
+    "babel-preset-amex": "^3.5.0",
+    "rimraf": "^2.5.2"
+  },
+  "one-amex": {
+    "bundler": {
+      "requiredExternals": [
+        "semver"
```
10xLaCroixDrinker, line 27: This should be installed as a dep and actually used in the module.

prod-sample/sample-modules/unhealthy-frank/0.0.0/src/components/UnhealthyFrank.jsx
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import React from 'react';
+
+export default function HealthyFrank() {
```
10xLaCroixDrinker, line 19:
```suggestion
export default function UnhealthyFrank() {
```

prod-sample/sample-modules/unhealthy-frank/0.0.0/README.md
10xLaCroixDrinker, line 1:
```suggestion
# unhealthy-frank
```

prod-sample/sample-modules/unhealthy-frank/0.0.0/README.md
```diff
@@ -0,0 +1,3 @@
...
+
+A module that works, nothing more, nothing less. Used for integration testing.
```
10xLaCroixDrinker, line 3:
```suggestion
A module that requires a missing external, nothing more, nothing less. Used for integration testing.
```

__tests__/integration/one-app.spec.js
```diff
@@ -700,30 +747,71 @@ describe('Tests that require Docker setup', () => {
...
+          test('fails to get external `semver` for child module as an unsupplied `requiredExternal` for new module in mooduleMap', async () => {
+            const revertErrorMatch = /There was an error loading module (?<moduleName>.*) at (?<url>.*). Ignoring (?<ignoredModule>.*) until .*/;
+            const requiredExternalsError = searchForNextLogMatch(revertErrorMatch);
+            const modName = 'unhealthy-frank';
+            const modVersion = '0.0.0';
+            await addModuleToModuleMap({
+              moduleName: modName,
+              version: modVersion,
+            });
+            // not ideal but need to wait for app to poll;
```
10xLaCroixDrinker, line 802: Please set a lower max poll time instead of waiting a full 5s

10xLaCroixDrinker (changes requested):

__tests__/integration/one-app.spec.js
JAdshead: 5 seconds is hardcoded as the min - https://github.com/americanexpress/one-app/blob/main/src/server/utils/pollModuleMap.js#L35-L40

JAdshead (approved):

__tests__/integration/one-app.spec.js
JAdshead: https://github.com/americanexpress/one-app/blob/main/src/server/utils/pollModuleMap.js#L35-L40

JAdshead: also for future reference here is where we set it - https://github.com/americanexpress/one-app/blob/main/prod-sample/one-app/base.env#L14

JAdshead:
Approved from what you have but we will need to fix those other tests

10xLaCroixDrinker (approved):

bishnubista (approved):

PR #640, opened 2021-12-30
Feat/disable csp 

dogpatch626:
This is to allow one-app to run modules without a valid csp in development mode. 
## Description

added function checkCsp to onModuleLoad to handle if statements regarding the presence of a newly introduced environment variable named ONE_DANGER_DISABLE_CSP. As well as this if the environment variable is present the header is not set with "content-security-policy" in the file csp.js. 
Note: this will be a draft so README will be updated before official merge. 
## Motivation and Context

This was needed to allow development of modules without having csp in development 

## How Has This Been Tested?

created test cases to assure if statements are tested, and ran locally.

when testing make sure to unset ONE_DANGER_DISABLE_CSP; 

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

- [x] This change impacts HTTP headers.


src/server/middleware/csp.js
```diff
@@ -69,7 +69,10 @@ const csp = () => (req, res, next) => {
   }
 
   res.scriptNonce = scriptNonce;
-  res.setHeader('Content-Security-Policy', updatedPolicy);
+
+  if (!process.env.ONE_DANGER_DISABLE_CSP) {
```
Matthew-Mallimo, line 73: Shouldn't we also check NODE_ENV ? The way it is currently CSP can be disabled in prod

src/server/utils/onModuleLoad.js
```diff
@@ -64,6 +64,18 @@ function validateConfig(configValidators, config) {
 
 export const CONFIGURATION_KEY = 'appConfig';
 
+export function cspCheck(csp) {
+  if (!csp && !process.env.ONE_DANGER_DISABLE_CSP) {
+    throw new Error('Root module must provide a valid content security policy. If you would like to bypass this, set the ONE_DANGER_DISABLE_CSP enviornment variable to *, and NODE_ENV to development.');
```
Matthew-Mallimo, line 69: Should we only display this message in development ? Maybe add some logic to only add the second sentence "If you would like to bypass this, set the ONE_DANGER_DISABLE_CSP environment variable to *, and NODE_ENV to development." if NODE_ENV='development'.

```suggestion
    throw new Error('Root module must provide a valid content security policy. If you would like to bypass this, set the ONE_DANGER_DISABLE_CSP environment variable to *, and NODE_ENV to development.');
```

src/server/utils/onModuleLoad.js
```diff
@@ -108,9 +121,7 @@ export default function onModuleLoad({
   }
 
   if (moduleName === serverStateConfig.rootModuleName) {
-    if (!csp) {
-      throw new Error('Root module must provide a valid content security policy');
-    }
+    cspCheck(csp);
```
Matthew-Mallimo, line 124: I don't love this floating function call. Maybe it should return true/false, and we keep the original if check ? Just personal preference, feel free to disagree.

Matthew-Mallimo (changes requested):

src/server/middleware/csp.js
dogpatch626: I put it as such because if the environment variable is present but NODE_ENV is not in dev the runner won't continue past the second check in checkCSP

src/server/utils/onModuleLoad.js
dogpatch626: I actually kind of agree, functionally it would be the same, But leaving this as is, is not the cleanliest. But what I will say is that it reduces a spot for complexity that could potentially be used later?

src/server/utils/onModuleLoad.js
dogpatch626: I think this makes sense, makes the console less cluttered

src/server/utils/onModuleLoad.js
10xLaCroixDrinker: I could go either way, but in either case this function needs a better name. Something along the lines of `validateCspIsPresent` if throwing here, and `isCspPresent` if returning a bool

__tests__/server/utils/onModuleLoad.spec.jsx
```diff
@@ -59,6 +60,7 @@ jest.mock('../../../src/server/middleware/sendHtml.js', () => ({
 
 const RootModule = () => <h1>Hello, world</h1>;
 const csp = "default: 'none'";
+const missingCsp = '';
```
10xLaCroixDrinker, line 63: Wouldn't the more common case here be that it is undefined?

src/server/utils/onModuleLoad.js
```diff
@@ -64,6 +64,18 @@ function validateConfig(configValidators, config) {
 
 export const CONFIGURATION_KEY = 'appConfig';
 
+export function cspCheck(csp) {
+  if (!csp && !process.env.ONE_DANGER_DISABLE_CSP) {
+    throw new Error('Root module must provide a valid content security policy. If you would like to bypass this, set the ONE_DANGER_DISABLE_CSP enviornment variable to *, and NODE_ENV to development.');
+  }
+  if (process.env.ONE_DANGER_DISABLE_CSP && process.env.NODE_ENV !== 'development') {
```
10xLaCroixDrinker, line 71: Assert on the value of the env var rather than the presence

Do this validation in [`runTime.js`](https://github.com/americanexpress/one-app/blob/main/src/server/config/env/runTime.js)

src/server/utils/onModuleLoad.js
```diff
@@ -98,6 +110,7 @@ export default function onModuleLoad({
   }
 
   const serverStateConfig = getServerStateConfig();
+
```
10xLaCroixDrinker, line 113: Why add this?

__tests__/server/middleware/csp.spec.js
```diff
@@ -62,6 +62,16 @@ describe('csp', () => {
       expect(headers).not.toHaveProperty('content-security-policy-report-only');
     });
 
+    it('does not set csp header if one_danger_disable_csp is present', () => {
```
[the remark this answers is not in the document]
10xLaCroixDrinker:
```suggestion
    it('does not set csp header if ONE_DANGEROUSLY_DISABLE_CSP is present', () => {
```

would keep env var names consistent

src/server/middleware/csp.js
10xLaCroixDrinker: Consider separating `checkCsp` into a util function that can be reused here

10xLaCroixDrinker (changes requested):

src/server/utils/onModuleLoad.js
dogpatch626: Um, I think this is just a mistake haha that's odd?

src/server/utils/onModuleLoad.js
dogpatch626: are you talking about doing something like 
```js 
{
    name: 'ONE_DANGER_DISABLE_CSP',
    normalize: (input) => input && input.toLowerCase(),
    valid: ['true', 'false'],
    defaultValue: 'false',
  },
  ```

Comment (10xLaCroixDrinker) [reply src/server/utils/onModuleLoad.js:71]: You can normalize to a bool and check for `NODE_ENV=development`  in `validate`. This lets the app fail to start before loading modules

File: __tests__/server/config/env/runTime.spec.js [hunk 8]
```diff
@@ -14,6 +14,7 @@
  * permissions and limitations under the License.
  */
 
+
```
Matthew-Mallimo, line 17:
```suggestion
```

__tests__/server/middleware/csp.spec.js
```diff
@@ -62,6 +63,16 @@ describe('csp', () => {
...
     });
 
+    it('does not set csp header if ONE_DANGER_DISABLE_CSP is present', () => {
+      process.env.ONE_DANGER_DISABLE_CSP = 'true';
+      const cspMiddleware = requireCSP().default;
+      cspMiddleware()(req, res, next);
+      // eslint-disable-next-line no-underscore-dangle
+      const headers = res._getHeaders();
+      expect(headers).not.toHaveProperty('content-security-policy');
+      process.env.ONE_DANGER_DISABLE_CSP = 'false';
```
Matthew-Mallimo, line 73:
```suggestion
```
This shouldn't be needed since `beforeEach` also sets this to false

Matthew-Mallimo (changes requested):

__tests__/server/utils/__snapshots__/onModuleLoad.spec.jsx.snap
```diff
@@ -12,4 +12,4 @@ exports[`onModuleLoad throws if the root module provides an incompatible version
 
 exports[`onModuleLoad throws when the one app version is incompatible 1`] = `"some-module@1.0.2 is not compatible with this version of one-app (4.43.0-0), it requires ~4.41.0."`;
 
-exports[`onModuleLoad validates csp added to the root module 1`] = `"Root module must provide a valid content security policy"`;
+exports[`onModuleLoad validates csp added to the root module 1`] = `"Cannot read property 'version' of undefined"`;
```
Matthew-Mallimo, line 15: This snapshot is incorrect

Matthew-Mallimo (changes requested):

__tests__/server/utils/__snapshots__/onModuleLoad.spec.jsx.snap
dogpatch626: Yea I just changed it back, it wasn't like that before I don't know why my snapshot is freaking out

Matthew-Mallimo (dismissed):

src/server/config/env/runTime.js
```diff
@@ -43,6 +43,24 @@ const runTime = [
     valid: ['development', 'production'],
     defaultValue: 'production',
   },
+  {
+    name: 'ONE_DANGER_DISABLE_CSP',
+    normalize: (input) => {
+      const parsed = input.toLowerCase();
```
10xLaCroixDrinker, line 49:
```suggestion
    normalize: (input) => {
      try {
        return JSON.parse(input.toLowerCase());
      } catch(e) {
        return input;
      }
    },
    validate: (input) => {      
```

Keep the validation separate from the normalization

Using `JSON.parse` like this will allow you to use an actual bool instead of the string in your assertions

src/server/config/env/runTime.js
```diff
@@ -43,6 +43,24 @@ const runTime = [
     valid: ['development', 'production'],
     defaultValue: 'production',
   },
+  {
+    name: 'ONE_DANGER_DISABLE_CSP',
```
10xLaCroixDrinker, line 47:
```suggestion
    name: 'ONE_DANGEROUSLY_DISABLE_CSP',
```

src/server/middleware/csp.js
```diff
@@ -70,7 +70,7 @@ const csp = () => (req, res, next) => {
 
   res.scriptNonce = scriptNonce;
 
-  if (!process.env.ONE_DANGER_DISABLE_CSP) {
+  if (process.env.ONE_DANGER_DISABLE_CSP === 'false') {
```
10xLaCroixDrinker, line 73:
```suggestion
  if (process.env.ONE_DANGER_DISABLE_CSP !== 'true') {
```

This assertion is a bit safer. If somehow there was an invalid value, the header would still be included

src/server/utils/onModuleLoad.js
```diff
@@ -64,15 +64,9 @@ function validateConfig(configValidators, config) {
...
-  if (!csp && !process.env.ONE_DANGER_DISABLE_CSP) {
-    throw new Error('Root module must provide a valid content security policy. If you would like to bypass this, set the ONE_DANGER_DISABLE_CSP enviornment variable to *, and NODE_ENV to development.');
-  }
-  if (process.env.ONE_DANGER_DISABLE_CSP && process.env.NODE_ENV !== 'development') {
-    throw new Error('If you are trying to bypass csp requirment, NODE_ENV must also be set to development.');
-  }
-  if (process.env.ONE_DANGER_DISABLE_CSP && process.env.NODE_ENV === 'development') {
-    console.warn('ONE_DANGER_DISABLE_CSP present and NODE_ENV set to development. Csp Header will not be set.');
+export function validateCspIsPresent(csp) {
+  if (!csp && process.env.ONE_DANGER_DISABLE_CSP === 'false') {
```
10xLaCroixDrinker, line 68:
```suggestion
  if (!csp && process.env.ONE_DANGER_DISABLE_CSP !== 'true') {
```

src/server/config/env/runTime.js
```diff
@@ -43,6 +43,24 @@ const runTime = [
...
+    normalize: (input) => {
+      const parsed = input.toLowerCase();
+      if (parsed === 'true' && process.env.NODE_ENV !== 'development') {
+        throw new Error('If you are trying to bypass csp requirment, NODE_ENV must also be set to development.');
```
10xLaCroixDrinker, line 51:
```suggestion
        throw new Error('If you are trying to bypass CSP requirement, NODE_ENV must also be set to development.');
```

src/server/config/env/runTime.js
```diff
@@ -43,6 +43,24 @@ const runTime = [
...
+      }
+      if (parsed === 'true' && process.env.NODE_ENV === 'development') {
+        console.warn('ONE_DANGER_DISABLE_CSP present and NODE_ENV set to development. Csp Header will not be set.');
```
10xLaCroixDrinker, line 54:
```suggestion
        console.warn('ONE_DANGEROUSLY_DISABLE_CSP is true and NODE_ENV is set to development. Content-Security-Policy header will not be set.');
```

__tests__/server/utils/onModuleLoad.spec.jsx
```diff
@@ -443,21 +444,6 @@ describe('onModuleLoad', () => {
...
-    delete process.env.ONE_DANGER_DISABLE_CSP;
-  });
-  it('Warns that csp header will not be set because ONE_DANGER_DISABLE_CSP is present and NODE_ENV is in development', () => {
-    process.env.ONE_DANGER_DISABLE_CSP = '*';
-    process.env.NODE_ENV = 'development';
-    cspCheck(missingCsp);
-    expect(consoleWarnSpy).toHaveBeenCalledTimes(1);
-    delete process.env.ONE_DANGER_DISABLE_CSP;
-    delete process.env.NODE_ENV;
+    expect(() => validateCspIsPresent(missingCsp)).toThrow('Root module must provide a valid content security policy.');
```
10xLaCroixDrinker, line 447: Can we include assertions for the other cases?

- CSP is set
- CSP is not set and the env var is

10xLaCroixDrinker (changes requested):

src/server/config/env/runTime.js
dogpatch626: Hmm, I wonder though. Does it even matter? because you can't save a bool as an env variable? I could be wrong. or is this just to better parse any input

dogpatch626: Im not too sure about this, in testing it returns a bool  rather than a string. Would this cause problems or does it not matter because it will be saved as a string in env anyways?

JAdshead: an issue with checking against a string value `"true"`  is that it increases the risk that a future us will break conditional in a refactor as its common in JS to check that something is truthy rather than true. 

someone may do this refactor and it could slide possibly slide through:
```
if (parsed && process.env.NODE_ENV !== 'development') {
//...
}
```
the issue being any populated string is truthy even `"false"` 
it is a small risk but possible.

__tests__/server/utils/onModuleLoad.spec.jsx
dogpatch626: well this gets tested in csp.spec.js as such 
```js it('sets a csp header in development', () => {
      process.env.NODE_ENV = 'development';
      const cspMiddleware = requireCSP().default;
      cspMiddleware()(req, res, next);
      // eslint-disable-next-line no-underscore-dangle
      const headers = res._getHeaders();
      expect(headers).toHaveProperty('content-security-policy');
      expect(headers).not.toHaveProperty('content-security-policy-report-only');
    });

    it('does not set csp header if ONE_DANGEROUSLY_DISABLE_CSP is present', () => {
      process.env.ONE_DANGEROUSLY_DISABLE_CSP = 'true';
      const cspMiddleware = requireCSP().default;
      cspMiddleware()(req, res, next);
      // eslint-disable-next-line no-underscore-dangle
      const headers = res._getHeaders();
      expect(headers).not.toHaveProperty('content-security-policy');
    });
```
other files are managing the lack of csp so Im not sure what exactly I would be testing for within onModuleLoad

__tests__/server/utils/onModuleLoad.spec.jsx
```diff
@@ -59,11 +60,13 @@ jest.mock('../../../src/server/middleware/sendHtml.js', () => ({
 
 const RootModule = () => <h1>Hello, world</h1>;
 const csp = "default: 'none'";
+const missingCsp = undefined;
 describe('onModuleLoad', () => {
   const consoleInfoSpy = jest.spyOn(console, 'info').mockImplementation(() => null);
   const consoleWarnSpy = jest.spyOn(console, 'warn').mockImplementation(() => null);
 
   beforeEach(() => {
+    process.env.ONE_DANGER_DISABLE_CSP = 'false';
```
10xLaCroixDrinker, line 69:
```suggestion
    process.env.ONE_DANGEROUSLY_DISABLE_CSP = false;
```

__tests__/server/utils/onModuleLoad.spec.jsx
```diff
@@ -440,4 +443,7 @@ describe('onModuleLoad', () => {
 
     expect(setConfigureRequestLog).toHaveBeenCalledWith(configureRequestLog);
   });
+  it('Throws error if csp and ONE_DANGER_DISABLE_CSP is not set', () => {
```
10xLaCroixDrinker, line 446:
```suggestion
  it('Throws error if csp and ONE_DANGEROUSLY_DISABLE_CSP is not set', () => {
```

__tests__/server/utils/onModuleLoad.spec.jsx
10xLaCroixDrinker: They are different conditions though. One could break one and not the other.

10xLaCroixDrinker (changes requested):

__tests__/server/config/env/runTime.spec.js
```diff
@@ -125,6 +127,36 @@ describe('runTime', () => {
...
 
+  describe('ONE_DANGEROUSLY_DISABLE_CSP', () => {
+    const disableCspEnv = getEnvVarConfig('ONE_DANGEROUSLY_DISABLE_CSP');
+
+    it('throws error if ONE_DANGEROUSLY_DISABLE_CSP is set to true and NODE_ENV is not in development', () => {
+      expect(() => disableCspEnv.validate('true')).toThrowError('If you are trying to bypass CSP requirement, NODE_ENV must also be set to development.');
+    });
+
+    it('warns console if both ONE_DANGEROUSLY_DISABLE_CSP and NODE_ENV are set properly', () => {
+      process.env.NODE_ENV = 'development';
```
Matthew-Mallimo, line 138: This never gets reset, and the `beforeEach` isn't there, was it removed ?

Matthew-Mallimo (changes requested):

__tests__/server/config/env/runTime.spec.js
dogpatch626: this was never in before each it just has a call to reset the NODE_ENV env var

dogpatch626: resetEnvVar('NODE_ENV'); <- this will delete that env var

Matthew-Mallimo: Where is this called ?

dogpatch626:
```js
beforeEach(() => {
    // eslint-disable-next-line no-console
    console.info = jest.fn();
    console.warn = jest.fn();
    resetEnvVar('NODE_ENV');
    resetEnvVar('ONE_DANGEROUSLY_DISABLE_CSP', 'false');
    resetEnvVar('HTTP_ONE_APP_DEV_CDN_PORT');
    jest.resetModules();
    jest.resetAllMocks();
  });```

**Review by Matthew-Mallimo — APPROVED** [review]

**Review by Matthew-Mallimo — APPROVED** [review]

File: docs/api/server/Environment-Variables.md [hunk 21]
```diff
@@ -536,6 +537,30 @@ ONE_DANGEROUSLY_ACCEPT_BREAKING_EXTERNALS=undefined
 ```
 
 
+## `ONE_DANGEROUSLY_DISABLE_CSP`
+
+**Runs In**
+* 🚫 Production
+* ✅ Development
+
+If set to `true`, one-app will not throw an error when a valid csp is not present. This flag is meant to allow any module to be ran as a root module and also allow root modules to bypass the csp requirement so long as NODE_ENV is in development. So long as NODE_ENV is in development and ONE_DANGEROUSLY_DISABLE_CSP is true a csp header will not be set.
```
10xLaCroixDrinker, line 546:
```suggestion
If set to `true`, one-app will not throw an error when a valid Content Security Policy (CSP) is not present. This flag is meant to allow any module to be ran as a root module and to allow root modules to bypass the CSP requirement so long as `NODE_ENV` is `development`. When `NODE_ENV` is `development` and `ONE_DANGEROUSLY_DISABLE_CSP` is `true` a CSP header will not be set. 
```

__tests__/server/config/env/runTime.spec.js
```diff
@@ -125,6 +127,36 @@ describe('runTime', () => {
     });
   });
 
+  describe('ONE_DANGEROUSLY_DISABLE_CSP', () => {
+    const disableCspEnv = getEnvVarConfig('ONE_DANGEROUSLY_DISABLE_CSP');
+
+    it('throws error if ONE_DANGEROUSLY_DISABLE_CSP is set to true and NODE_ENV is not in development', () => {
```
10xLaCroixDrinker, line 133:
```suggestion
    it('throws error if ONE_DANGEROUSLY_DISABLE_CSP is set to true and NODE_ENV is not development', () => {
```

__tests__/server/config/env/runTime.spec.js
```diff
@@ -125,6 +127,36 @@ describe('runTime', () => {
...
+    });
+
+    it('parses input and returns it in lowercase', () => {
+      expect(disableCspEnv.normalize('TRUE')).toBe('true');
+    });
+
+    it('returns the original string if not parsable', () => {
+      expect(disableCspEnv.normalize(1)).toBe(1);
+    });
+
```
10xLaCroixDrinker, line 154:
```suggestion
```

src/server/config/env/runTime.js
```diff
@@ -43,6 +43,30 @@ const runTime = [
...
   },
+  {
+    name: 'ONE_DANGEROUSLY_DISABLE_CSP',
+    normalize: (input) => {
+      try {
+        return input.toLowerCase();
+      } catch (e) {
+        return input;
+      }
+    },
```
10xLaCroixDrinker, line 54:
```suggestion
    normalize: (input) => input.toLowerCase(),
```

src/server/config/env/runTime.js
```diff
@@ -43,6 +43,30 @@ const runTime = [
...
+      if (input === 'true' && process.env.NODE_ENV !== 'development') {
+        throw new Error('If you are trying to bypass CSP requirement, NODE_ENV must also be set to development.');
+      }
+      if (input === 'true' && process.env.NODE_ENV === 'development') {
+        console.warn('ONE_DANGEROUSLY_DISABLE_CSP is true and NODE_ENV is set to development. Content-Security-Policy header will not be set.');
+      }
+      if (input === 'false') {
+        return input;
+      }
+      return input;
```
10xLaCroixDrinker, line 65:
```suggestion
```

10xLaCroixDrinker (changes requested):

docs/api/server/Environment-Variables.md
```diff
@@ -536,6 +537,30 @@ ONE_DANGEROUSLY_ACCEPT_BREAKING_EXTERNALS=undefined
...
+
+**Runs In**
+* 🚫 Production
+* ✅ Development
+
+If set to `true`, one-app will not throw an error when a valid Content Security Policy (CSP) is not present. This flag is meant to allow any module to be ran as a root module and to allow root modules to bypass the CSP requirement so long as `NODE_ENV` is `development`. When `NODE_ENV` is `development` and `ONE_DANGEROUSLY_DISABLE_CSP` is `true` a CSP header will not be set. 
+
+**Shape**
+```bash
+ONE_DANGEROUSLY_DISABLE_CSP=true
```
JAdshead, line 550:
```suggestion
ONE_DANGEROUSLY_DISABLE_CSP=Boolean
```

JAdshead (approved):

JAdshead (changes requested):
just the one change

JAdshead (approved):

10xLaCroixDrinker (approved):

PR #686, opened 2022-02-16
Revert "v5 on Node 16"

Matthew-Mallimo:
Reverts americanexpress/one-app#677


10xLaCroixDrinker:
Do we need to revert all the changes or could we just do the Dockerfile itself?

Matthew-Mallimo:
> Do we need to revert all the changes or could we just do the Dockerfile itself?

Not sure, were the additional changes made because of the node 16 change ?

PR #711, opened 2022-04-07
chore(eslint): update

JAdshead:
## Description

Update eslint and eslint-config-amex to resolve peer deps conflicts

## Types of Changes

- [x] Dependency update
- [ ] Security update

## Checklist:


__tests__/client/prerender.spec.js
```diff
@@ -134,15 +133,15 @@ describe('loadPrerenderScripts', () => {
...
+    [...document.querySelectorAll('script')].forEach((scriptTag) => { scriptTag.remove(); });
 
     jest.spyOn(document, 'addEventListener').mockImplementation(() => { /* noop */ });
   });
 
   const createScript = ({ src, helmet }) => {
     const script = document.createElement('script');
     script.src = src;
-    if (helmet) script.setAttribute('data-react-helmet', true);
+    if (helmet) script.dataset.reactHelmet = true;
```
Matthew-Mallimo, line 144: why did this change?

__tests__/client/prerender.spec.js
```diff
@@ -161,7 +160,7 @@ describe('moveHelmetScripts', () => {
       { src: 'baz.js', helmet: true },
     ]
       .map(createScript)
-      .forEach((script) => document.body.appendChild(script));
+      .forEach((script) => document.body.append(script));
```
Matthew-Mallimo, line 163: Was this preferred by the linter?

dogpatch626 (approved):

__tests__/server/devHolocronCDN.spec.js
```diff
@@ -109,51 +112,38 @@ describe('devHolocronCDN', () => {
...
-          expect(corsMiddleware).toHaveBeenCalledTimes(1);
-          expect(oneAppDevCdnMiddleware).not.toHaveBeenCalled();
-          expect(status).toBe(204);
-          done();
-        });
+      const resp = await request(devHolocronCDN)
+        .get('/static/anything.json');
+      expect(corsMiddleware).toHaveBeenCalledTimes(1);
+      expect(oneAppDevCdnMiddleware).not.toHaveBeenCalled();
+      expect(resp.status).toBe(204);
```
Matthew-Mallimo, line 132: since this is async we should add an `expect.assertions`

__tests__/server/devHolocronCDN.spec.js
```diff
@@ -109,51 +112,38 @@ describe('devHolocronCDN', () => {
...
-        .end(() => {
-          expect(corsMiddleware).toHaveBeenCalledTimes(1);
-          expect(oneAppDevCdnMiddleware).toHaveBeenCalledTimes(1);
-          done();
-        });
+    it('should hit the one-app-dev-cdn middleware after cors', async () => {
+      await request(devHolocronCDN)
+        .get('/static/anything.json');
+      expect(corsMiddleware).toHaveBeenCalledTimes(1);
+      expect(oneAppDevCdnMiddleware).toHaveBeenCalledTimes(1);
```
Matthew-Mallimo, line 139:
```suggestion
      expect.assertions(2);
      await request(devHolocronCDN)
        .get('/static/anything.json');
      expect(corsMiddleware).toHaveBeenCalledTimes(1);
      expect(oneAppDevCdnMiddleware).toHaveBeenCalledTimes(1);
```

__tests__/server/devHolocronCDN.spec.js
```diff
@@ -109,51 +112,38 @@ describe('devHolocronCDN', () => {
...
-        .end(() => {
-          expect(corsMiddleware).toHaveBeenCalledTimes(1);
-          expect(oneAppDevCdnMiddleware).not.toHaveBeenCalled();
-          done();
-        });
+    it('should miss the one-app-dev-cdn middleware if not a static route', async () => {
+      await request(devHolocronCDN)
+        .get('/not-static.json');
+      expect(corsMiddleware).toHaveBeenCalledTimes(1);
+      expect(oneAppDevCdnMiddleware).not.toHaveBeenCalled();
```
Matthew-Mallimo, line 146:
```suggestion
			expect.assertions(2);
      await request(devHolocronCDN)
        .get('/not-static.json');
      expect(corsMiddleware).toHaveBeenCalledTimes(1);
      expect(oneAppDevCdnMiddleware).not.toHaveBeenCalled();
```

__tests__/server/listen.spec.js
```diff
@@ -113,24 +113,22 @@ describe('server listen', () => {
       expect(cb).toHaveBeenCalledWith(null, { port: '8998' });
     });
 
-    it('passes the error to the callback', (done) => {
+    it('passes the error to the callback', async () => {
       process.env.HTTP_PORT = 8998;
       const listenError = new Error('port taken or such');
       http.mock.listenError = listenError;
-      listenHttp(app, (err, serverInfo) => {
+      await listenHttp(app, (err, serverInfo) => {
```
Matthew-Mallimo, line 120: Should add `expect.assertions(2);`

__tests__/server/listen.spec.js
```diff
@@ -113,24 +113,22 @@ describe('server listen', () => {
...
-        done();
       });
     });
 
-    it('passes the server information to the callback', (done) => {
+    it('passes the server information to the callback', async () => {
       process.env.HTTP_PORT = 8998;
       http.mock.listenError = null;
-      listenHttp(app, (err, serverInfo) => {
+      await listenHttp(app, (err, serverInfo) => {
```
Matthew-Mallimo, line 129: should add `expect.assertions(2);`

docs/guides/Code-Splitting-Using-Holocron.md
```diff
@@ -212,7 +212,7 @@ MyHolocronModule.holocron = {
       const imported = await import(/* webpackChunkName: '<chunkName>' */ './Chunk');
       Chunk = imported.default || imported;
     } catch (error) {
-      Chunk = () => <p>{error.message}</p>;
+      Chunk = function ChunkError() { return <p>{error.message}</p>; };
```
Matthew-Mallimo, line 215: why didn't it like the arrow function in a markdown file?

src/client/badPartMonkeypatches.js
```diff
@@ -48,7 +48,7 @@ if (typeof document !== 'undefined') {
     if (typeof cb === 'string') {
       throw new TypeError(`eval form of ${name} used, this is disabled`);
     }
-    return orig.apply(this, [cb, ...args]);
+    return Reflect.apply(orig, this, [cb, ...args]);
```
Matthew-Mallimo, line 51: TIL [Reflect](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect) is a global object

Matthew-Mallimo (changes requested):
Few additional test cases that I didn't comment on (to prevent spam) that would need `expect.assertions`

__tests__/client/prerender.spec.js
JAdshead: yes https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-dom-node-append.md

__tests__/server/listen.spec.js
JAdshead: you only need expect.assertions for conditional test

docs/guides/Code-Splitting-Using-Holocron.md
JAdshead: it creates an unnamed component

__tests__/server/devHolocronCDN.spec.js
JAdshead: not required due to async await pattern

10xLaCroixDrinker: not tracking why this shouldn't be required for all a sync tests

__tests__/server/utils/batchModulesToUpdate.spec.js
```diff
@@ -54,8 +60,9 @@ describe('batchModulesToUpdate', () => {
     const result = batchModulesToUpdate(modulesWithoutRoot);
     expect(result.length).toBe(7);
     result.forEach((arr, i) => {
-      if (i < 6) expect(arr.length).toBe(10);
-      else if (i === 6) expect(arr.length).toBe(5);
+      // if (i < 6) expect(arr.length).toBe(10);
+      // else if (i === 6) expect(arr.length).toBe(5);
```
10xLaCroixDrinker, line 64: Remove commented code

package.json
```diff
@@ -151,9 +150,11 @@
     "danger": "^10.6.0",
     "enzyme": "^3.11.0",
     "enzyme-to-json": "^3.6.1",
-    "eslint": "^7.12.1",
-    "eslint-config-amex": "^11.2.0",
+    "eslint": "^7.32.0",
+    "eslint-config-amex": "^14.2.1",
```
10xLaCroixDrinker, line 154: Why aren't we going to 15?

prod-sample/sample-modules/frank-lloyd-root/0.0.0/src/components/FrankLloydRoot.jsx
```diff
@@ -41,12 +41,12 @@ export function FrankLloydRoot({ children, config }) {
 }
 
 FrankLloydRoot.childRoutes = (store) => [
-  <ModuleRoute path="vitruvius" moduleName="vitruvius-franklin" />,
-  <ModuleRoute path="success" component={HelloWorldComponent} />,
-  <ModuleRoute path="healthy-frank" moduleName="healthy-frank" />,
-  <ModuleRoute path="demo/:moduleName" moduleName="preview-frank" />,
-  <ModuleRoute path="html-partial/:locale/:moduleName" component={Partial} onEnter={onPartialRouteEnter(store, true)} />,
-  <ModuleRoute path="text-only/:locale/:moduleName" component={Partial} onEnter={onPartialRouteEnter(store, false, true)} />,
+  <ModuleRoute key="vitruvius" path="vitruvius" moduleName="vitruvius-franklin" />,
```
10xLaCroixDrinker, line 44: Could we just not use an array instead of cluttering with all these keys?

src/client/service-worker/events/utility.js
```diff
@@ -20,7 +20,8 @@ import {
 
 // language
 const appLocaleRegExp = /i18n\/([^/]*)\.js$/;
-const moduleLocaleRegExp = /([a-z]{2,3}(-[a-zA-Z]{1,})?)\/[^/]*\.json$/;
+// eslint-disable-next-line unicorn/no-unsafe-regex
```
10xLaCroixDrinker, line 23: Can we make this safe?

src/server/metrics/app-version.js
```diff
@@ -29,6 +29,7 @@ createGauge({
 
 // keep the parsing in a local variable to avoid keeping it in memory forever
 function parseVersionAndSetGaugue() {
+  // eslint-disable-next-line unicorn/no-unsafe-regex
```
10xLaCroixDrinker, line 32: Can we make this safe?

__tests__/client/initClient.spec.js
```diff
@@ -21,6 +21,7 @@ import { fromJS } from 'immutable';
 jest.mock('react', () => {
   const StrictMode = ({ children }) => children;
   const react = jest.requireActual('react');
+  // eslint-disable-next-line jest/no-standalone-expect
```
10xLaCroixDrinker, line 24: Is this expect needed?

10xLaCroixDrinker (changes requested):

__tests__/server/devHolocronCDN.spec.js
JAdshead: expect.assertions is used to ensure that X amount of expect are called in scenarios where its possible for the block to finish without calling the expect. This was very important when unhandled promises did not throw an error. 


in this case it used to be possible for the function to return without calling the expect. if the promise rejected this could silently finish without calling the expect which is why we there is the `expect.assertions`. 
```
it('has expect in promise chain', () => {
  somePromise().then((result) => {
    expect(result).toBe(true)
  })
})
```

Now however if `somePromise` fails it will throw 

```
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "error".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

JAdshead: `expect.assertions` is still relevant for conditional logic or where you are using a try catch block

```
it('in try catch block', () => {
    try {
      const wasSafe = possibleKaboom()
      expect(wasSafe).toBe(true)
    } catch {
      // if the above block throws an error this test can fail silently 
      // which is why we need expect.assertions.
    }
  })
```

```
it('conditional tests', () => {
    const itDid = doSomething()
    if(itDid) {
      expect(itDid).toBe(true)
    } else {
      // this is a bad test but if for some reason you need this pattern
      // then expect.assertions is required here
    }
  })
  ```

File: __tests__/server/middleware/csp.spec.js [hunk 16]
```diff
@@ -18,8 +18,10 @@ import httpMocks from 'node-mocks-http';
 
 const sanitizeCspString = (cspString) => cspString
 // replaces dynamic ip and nonce to prevent snapshot failures
-  .replace(/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/g, '0.0.0.0')
-  .replace(/nonce-[a-zA-Z0-9]{8}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{12}/, 'nonce-00000000-0000-0000-0000-000000000000');
+  // eslint-disable-next-line unicorn/no-unsafe-regex
```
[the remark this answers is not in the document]
JAdshead: probably but as these are test regexs its less critical.

__tests__/client/initClient.spec.js
JAdshead: i don't think it holds much value

__tests__/client/prerender.spec.js
JAdshead: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v23.0.0/docs/rules/prefer-dataset.md

package.json
JAdshead: 15 has another 200+ changes, most of which require a significant effort

10xLaCroixDrinker (approved):

Matthew-Mallimo (approved):

PR #740, opened 2022-05-11
chore(workflow): allow control of docker meta tag

code-forger:
This PR allows control over the docker meta tag (such as `latest` `next` etc)

1. Three workflows (Release 3, 4, and 5) now accept a required choice of `latest`, `next`, or `do_not_tag`.
- Although Release 3 does not use this value, it will pass it to Release 4. Since Release 3 is the 'entrypoint' for releasing arbitrary tags, Release 3 must receive this value and pass it on.

2. The `Normalize meta tag` step of Release 3, 4, and 5 will determine the value, in this order:
- If the previous workflow passed in a value, that will be used
- If the user has specified a value, that will be used
- For automatic runs on the main branch, the value will be `latest`
- For automatic runs on any other branch the value will be `do_not_tag`

3. The value is used while tagging and pushing the docker image (present in Release 4 and 5 only)
- If the value is `do_not_tag`, the meta tag will be skipped (the image will still be tagged with the versions as normal)
- if the value is `do_not_tag`, the major version tag will also be skipped. The major version tag acts like a `latest in this major stream` tag, and so if we are not tagging `latest` we should not tag the major either.
- For any other run, the tag will take the value as specified in `Normalize meta tag`

4. The value is passed forward to the next workflow (present in Release 3 and 4 only)

- Since Release 6 has no use for the value, Release 5 does not need to pass it on

## How Has This Been Tested?
Running workflows in a fork of this repo, including reading the logs from `Normalize meta tag` and seeing the tag used at the tag step

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/.github/workflows/release-step-3_automatic_prepare-docker-tags.yml
+++ b/.github/workflows/release-step-3_automatic_prepare-docker-tags.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Normalize meta tag
         id: set_docker_meta_tag
         run: |
-          # if a meta tag was specified this was a menual build
+          # if a meta tag was specified this was a manual build
           if [[ "${{ github.event.inputs.docker_meta_tag }}" != "" ]]
           then
           echo ::set-output name=metatag::${{ github.event.inputs.docker_meta_tag }}
--- a/.github/workflows/release-step-3_automatic_prepare-docker-tags.yml
+++ b/.github/workflows/release-step-3_automatic_prepare-docker-tags.yml
@@ -5,6 +5,14 @@ on:
     tag_version:
      description: 'Semantic Version'
      required: true
+    docker_meta_tag:
+     type: choice
+     options:
+      - do_not_tag
+      - latest
+      - next
+     description: When manually triggering a deployment, specify which meta tag such as `latest` should be used. `do_not_tag` will result in no meta tag being applied.
+     required: true
   push:
     tags:
       - '*'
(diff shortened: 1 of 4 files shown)


.github/workflows/release-step-3_automatic_prepare-docker-tags.yml
```diff
@@ -5,6 +5,14 @@ on:
     tag_version:
      description: 'Semantic Version'
      required: true
+    docker_meta_tag:
+     type: choice
+     options:
+      - do_not_tag
+      - latest
+      - next
```
JAdshead, line 13: what about setting a default ?

JAdshead (approved):

.github/workflows/release-step-3_automatic_prepare-docker-tags.yml
code-forger: it defaults to the first choice:
image

.github/workflows/release-step-3_automatic_prepare-docker-tags.yml
```diff
@@ -30,6 +38,24 @@ jobs:
           echo ::set-output name=tagversion::${{ github.event.inputs.tag_version }}
           fi
           echo $tagversion
+      - name: Normalize meta tag
+        id: set_docker_meta_tag
+        run: |
+          # if a meta tag was specified this was a menual build
```
Matthew-Mallimo, line 44:
```suggestion
          # if a meta tag was specified this was a manual build
```

.github/workflows/release-step-4_automatic_docker-prod-build-and-publish.yml
```diff
@@ -36,6 +44,29 @@ jobs:
...
           fi
+      - name: Normalize meta tag
+        id: set_docker_meta_tag
+        run: |
+          # Always use the value from the last step if it is present
+          if [[ "${{ github.event.client_payload.docker_meta_tag }}" != "" ]]
+          then
+          echo ::set-output name=metatag::${{ github.event.client_payload.docker_meta_tag }}
+          echo From previous step: ${{ github.event.client_payload.docker_meta_tag }}
+          # if a meta tag was specified this was a menual build
```
Matthew-Mallimo, line 55:
```suggestion
          # if a meta tag was specified this was a manual build
```

.github/workflows/release-step-5_automatic_docker-dev-build-and-publish.yml
```diff
@@ -28,6 +36,29 @@ jobs:
...
           fi
+      - name: Normalize meta tag
+        id: set_docker_meta_tag
+        run: |
+          # Always use the value from the last step if it is present
+          if [[ "${{ github.event.client_payload.docker_meta_tag }}" != "" ]]
+          then
+          echo ::set-output name=metatag::${{ github.event.client_payload.docker_meta_tag }}
+          echo From previous step: ${{ github.event.client_payload.docker_meta_tag }}
+          # if a meta tag was specified this was a menual build
```
Matthew-Mallimo, line 47:
```suggestion
          # if a meta tag was specified this was a manual build
```

10xLaCroixDrinker (approved):

PR #755, opened 2022-06-06
fix(renderPartial): respect disabledStyles for partial render

aaka5h:
## Description

Respect `disableStyles` to remove ssr-css for partial routes

## Motivation and Context

While disableStyles set via `setDangerouslyDisableScriptsAndStyles` when set for normal routes removes the ssr-css, but the disableStyles value is not respected in case of partial routes (set via `renderPartialOnly`/`setRenderPartialOnly` action)
  
## How Has This Been Tested?

Unit test added

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


10xLaCroixDrinker (approved):
thanks for the contribution!

Matthew-Mallimo (approved):

PR #802, opened 2022-08-04
feat: routes sanity checks

giulianok:
Adding some sanity checks for routes to make sure the right headers, response, and status code are in place

## Description

Added a new integration test case to check some of the routes' headers, response, and status code

## Motivation and Context

This makes the integration tests stronger and helps the fastify migration process to make sure everything is in place

## How Has This Been Tested?

Integration tests

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


__tests__/integration/one-app.spec.js
```diff
@@ -1668,3 +1668,213 @@ describe('heapdump', () => {
...
+    expect(rawHeaders['content-length']).toEqual(['2']);
+    expect(rawHeaders['content-type']).toEqual(['text/plain; charset=utf-8']);
+    expect(rawHeaders['referrer-policy']).toEqual(['same-origin']);
+    expect(rawHeaders['strict-transport-security']).toEqual([
+      'max-age=15552000; includeSubDomains',
+    ]);
+    expect(rawHeaders.vary).toEqual(['Accept-Encoding']);
+    expect(rawHeaders['x-content-type-options']).toEqual(['nosniff']);
+    expect(rawHeaders['x-frame-options']).toEqual(['DENY']);
+    expect(rawHeaders['x-xss-protection']).toEqual(['1; mode=block']);
```
10xLaCroixDrinker, line 1729: What about an inline snapshot of the headers?

10xLaCroixDrinker (approved):

__tests__/integration/one-app.spec.js
giulianok: we could use inline snapshot, only problem is that some header values are dynamic, like `date` and `etag`, maybe there's a way to mock them

10xLaCroixDrinker: check out asymmetric snapshots

__tests__/integration/one-app.spec.js
```diff
@@ -1668,3 +1705,323 @@ describe('heapdump', () => {
...
+        expect.any(String),
+      ],
+      etag: [
+        expect.any(String),
+      ],
+      'expect-ct': [
+        'max-age=0',
+      ],
+      'one-app-version': [
+        '6.0.0-0520af2e',
```
JAdshead, line 1865:
```suggestion
        expect.any(String),
```

__tests__/integration/one-app.spec.js
```diff
@@ -1668,3 +1705,323 @@ describe('heapdump', () => {
     expect(size).toBeLessThanOrEqual(100 * oneMegabyte);
   });
 });
+
+describe('Routes sanity checks', () => {
```
JAdshead, line 1709: these should be moved to be in the first block https://github.com/americanexpress/one-app/blob/main/__tests__/integration/one-app.spec.js#L59

10xLaCroixDrinker, line 1709:
```suggestion
describe('Routes confidence checks', () => {
```

The term "sanity check" is ableist

__tests__/integration/one-app.spec.js
```diff
@@ -1668,3 +1705,323 @@ describe('heapdump', () => {
...
+  test('Request: /_/report/errors responds with status 415', async () => {
+    const response = await fetch(`${fetchUrl}/_/report/errors`, {
+      ...defaultFetchOptions,
+      method: 'POST',
+      headers: {
+        origin: 'test.example.com',
+      },
+    });
+
+    expect(response.headers.raw()).toEqual({
```
10xLaCroixDrinker, line 1842: Do we need all the headers for these tests or can we use `expect.objectContaining` to target the most relevant ones?

10xLaCroixDrinker (changes requested):

__tests__/integration/one-app.spec.js
giulianok: we could, I thought the feedback about using inline snapshot (https://github.com/americanexpress/one-app/pull/802#discussion_r938331534) was about checking all the headers beside readability. In my opinion, I think it's good to check all the them, especially if a new one gets added in the future, the `toEqual` assertion will enforce the check of all headers including the new ones

10xLaCroixDrinker (approved):

JAdshead (approved):

10xLaCroixDrinker:
the commit type should have been test rather than feat

PR #803, opened 2022-08-10
refactor: migrating metrics server to Fastify

giulianok:
Migrating Metrics Server from ExpressJS to Fastify

## Description

This PR includes the migration of Metrics Server routes to Fastify + converting logging middleware into a fastify plugin

### Metrics Server

All routes have been moved to Fastify

### Logging

The logging middleware has been converted into a Fastify plugin. It required some refactoring in order to leverage fastify hooks.

To calculate `ttfb` (time to first byte), instead of monkey patching `writeHead` function from the raw/nodejs response, I leveraged the `onSend` fastify hook which is called right before `writeHead` is called, providing the same time (technically isn't the same but we are using Math.round when logging, so "it's the same").

Besides logging the full request duration and ttfb, I also added extra timers to calculate things like how long it took to start resolving a route, how long it took to resolved a route handler, etc.

Here's an example of a request that's taking 500ms to resolve things before fastify can start resolving the route handler, and the route handler takes 1000ms to resolved.

Screen Shot 2022-08-10 at 11 53 13 AM

I achieved this by adding a sleep function to create the delay

Screen Shot 2022-08-10 at 11 53 28 AM

Screen Shot 2022-08-10 at 11 53 33 AM

These are real metrics from `/metrics` and `/im-up`, where we can see that `/im-up` is a bit slower because of the healthcheck fn

`/metrics`
Screen Shot 2022-08-10 at 11 58 40 AM

`/im-up`
Screen Shot 2022-08-10 at 11 58 32 AM

As we can see, we do not have an overhead on the request (pre route handler) and on the response builder

## Motivation and Context

Allows us to run Metrics Server in Fastify without the need of Express

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


src/server/utils/logging/fastifyPlugin.js
```diff
@@ -0,0 +1,179 @@
...
+export const $RouteHandler = Symbol('$RouteHandler');
+export const $ResponseBuilder = Symbol('$ResponseBuilder');
+
+const TIMERS = {};
+const TENANT_UTILS = {
+  configureRequestLog: ({ log }) => log,
+};
+
+const getLocale = (req) => {
+  // TODO: Verify if `store` is available
```
giulianok, line 44: this only affects the locale in logger

src/server/metricsServer.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 American Express Travel Related Services Company, Inc.
+ * Copyright 2022 American Express Travel Related Services Company, Inc.
```
JAdshead, line 2: years in existing files can stay the same

src/server/metricsServer.js
```diff
@@ -14,49 +14,37 @@
...
-
-  fastify.express.disable('x-powered-by');
-  fastify.express.disable('e-tag');
+  await fastify.register(rateLimit, {
+    max: 120,
+    timeWindow: '1 minute',
+  });
+  await fastify.register(fp(logging));
+  await fastify.register(helmet);
+  await fastify.register(healthCheck);
```
JAdshead, line 37: i'm not sold on having the route abstracted into the plugin. I would have to do into each plugin to understand what routes this serves.

src/server/ssrServer.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 American Express Travel Related Services Company, Inc.
+ * Copyright 2022 American Express Travel Related Services Company, Inc.
```
JAdshead, line 2: this can be reverted

src/server/utils/logging/fastifyPlugin.js
```diff
@@ -0,0 +1,179 @@
+/*
+ * Copyright 2019 American Express Travel Related Services Company, Inc.
```
JAdshead, line 2: new file so should be this year.
```suggestion
 * Copyright 2022 American Express Travel Related Services Company, Inc.
```

src/server/utils/logging/fastifyPlugin.js
```diff
@@ -0,0 +1,179 @@
...
+
+Why symbols?
+It prevents conflicts with existing keys, native or from other plugins
+*/
+export const $RequestFullDuration = Symbol('$RequestFullDuration');
+export const $RequestOverhead = Symbol('$RequestOverhead');
+export const $RouteHandler = Symbol('$RouteHandler');
+export const $ResponseBuilder = Symbol('$ResponseBuilder');
+
+const TIMERS = {};
```
JAdshead, line 38: is this going to be thread safe ?

giulianok: just looking through the code, this is not really needed, i'll remove it

src/server/metricsServer.js
giulianok: that's true. We could pass the url through the options but still hides the route. Best option would be to create a reply decorator so we can implement the route outside the plugin

src/server/utils/logging/fastifyPlugin.js
giulianok: this is not longer a blocker, the Metrics Server does not inject the store, only the SSR Server which is being migrated in a separate PR

src/server/utils/logging/fastifyPlugin.js
```diff
@@ -0,0 +1,185 @@
...
+
+Why symbols?
+It prevents conflicts with existing keys, native or from other plugins
+*/
+export const $RequestFullDuration = Symbol('$RequestFullDuration');
+export const $RequestOverhead = Symbol('$RequestOverhead');
+export const $RouteHandler = Symbol('$RouteHandler');
+export const $ResponseBuilder = Symbol('$ResponseBuilder');
+
+const TENANT_UTILS = {
```
10xLaCroixDrinker, line 39: Rename

__tests__/server/plugins/healthCheck.spec.js
```diff
@@ -145,61 +140,119 @@ describe('healthCheck', () => {
...
+              0,
+              100,
+            ],
+          },
+        }
+      `);
     });
 
-    it('should return a 500 if it can\'t get the stats', async () => {
+    it("should return a 500 if it can't get the stats", async () => {
```
10xLaCroixDrinker, line 249: This didn't set off lint errors?

__tests__/server/ssrServer.spec.js
```diff
@@ -1,5 +1,6 @@
+/* eslint-disable jest/no-disabled-tests */
 /*
```
10xLaCroixDrinker, line 2:
```suggestion
```

__tests__/server/plugins/healthCheck.spec.js
giulianok: nope, ran `npm run test:lint -- --fix`

src/server/utils/logging/fastifyPlugin.js
giulianok: renamed to `UTILS` (`TENANT_UTILS` is the original name just fyi)

src/server/utils/logging/fastifyPlugin.js
```diff
@@ -0,0 +1,185 @@
...
+        requestOverhead: Math.round(getTimer($RequestOverhead)),
+        routeHandler: Math.round(getTimer($RouteHandler)),
+        responseBuilder: Math.round(getTimer($ResponseBuilder)),
+      },
+      statusCode: reply.statusCode,
+      statusText: reply.raw.statusMessage || undefined,
+    },
+  };
+
+  // TODO: 'req' is different from 'request', requires analysis
```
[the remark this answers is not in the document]
giulianok: just added more information in the comment, lmk

10xLaCroixDrinker (approved):

JAdshead (approved):

PR #821, opened 2022-09-02
chore(workflows): run unit tests on node 18

10xLaCroixDrinker:
## Description

Added node 18 to matrix for unit test action. Updated flags for build to support node 18

## Motivation and Context

This will help us prevent any regressions in node 18 while we continue to wait for it to become LTS

## How Has This Been Tested?

Ran `npm run build` and `npm run test:unit` in both node 16 and 18

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

Fixes build if using node 18


package.json
```diff
@@ -13,7 +13,9 @@
     "clean:build": "rimraf dist lib build lastBuild.json stats.json",
     "clean:test": "rimraf test-results",
     "prebuild": "npm run clean:build",
-    "build:bundle": "cross-env NODE_OPTIONS=--max-old-space-size=4096 bundle-one-app",
+    "build:bundle": "if-node-version \">16\" && npm run build:bundle:gt-16  || npm run build:bundle:16",
+    "build:bundle:gt-16": "cross-env \"NODE_OPTIONS=--max-old-space-size=4096 --openssl-legacy-provider\" bundle-one-app",
```
10xLaCroixDrinker, line 17: Using `--openssl-legacy-provider` in node < 17 fails because that node option is not supported until then, but this is required to resolve `webpack/webpack#14532`

Ultimately this should be resolved in the bundler itself, but this will get our unit test action working for now to prevent regressions in node 18

code-forger (approved):

PR #854, opened 2022-11-08
Converted small middlewares into plugins

giulianok:
## Description
Converted small ExpressJS middlewares into Fastify plugins. These are considered small since have relatively small code and they are (almost) completely abstract

## Motivation and Context

By converting ExpressJS middlewares into Fastify plugins, we can migrate routes from Express to Fastify to improve app's performance

## How Has This Been Tested?

Unit testing

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/__tests__/server/plugins/addCacheHeaders.spec.js
+++ b/__tests__/server/plugins/addCacheHeaders.spec.js
@@ -14,23 +14,48 @@
  * permissions and limitations under the License.
  */
 
-import addCacheHeaders from '../../../src/server/middleware/addCacheHeaders';
+import addCacheHeaders from '../../../src/server/plugins/addCacheHeaders';
 
-describe('addCacheHeaders', () => {
-  it('should add all expected cache headers', () => {
-    const req = { get: jest.fn(), headers: {} };
-    const res = { set: jest.fn((key, value) => value) };
-    const next = jest.fn();
-    addCacheHeaders(req, res, next);
+test('adds cache headers', () => {
+  const request = {
+    method: 'get',
+  };
+  const reply = {
+    header: jest.fn(),
+  };
+  const fastify = {
+    addHook: jest.fn(async (_hook, cb) => {
+      await cb(request, reply);
+    }),
+  };
+  const done = jest.fn();
 
-    const cacheHeaders = {
-      'Cache-Control': 'no-store',
-      Pragma: 'no-cache',
-    };
-    expect(res.set.mock.calls.length).toEqual(Object.keys(cacheHeaders).length);
-    Object.keys(cacheHeaders).forEach(
-      (header) => expect(res.set).toBeCalledWith(header, cacheHeaders[header])
-    );
-    expect(next).toBeCalled();
-  });
+  addCacheHeaders(fastify, null, done);
+
+  expect(fastify.addHook).toHaveBeenCalled();
+  expect(done).toHaveBeenCalled();
+  expect(reply.header).toHaveBeenCalledTimes(2);
+  expect(reply.header).toHaveBeenCalledWith('Cache-Control', 'no-store');
+  expect(reply.header).toHaveBeenCalledWith('Pragma', 'no-cache');
+});
+
+test('does not add cache headers', () => {
+  const request = {
+    method: 'post',
+  };
+  const reply = {
+    header: jest.fn(),
+  };
+  const fastify = {
+    addHook: jest.fn(async (_hook, cb) => {
+      await cb(request, reply);
+    }),
+  };
+  const done = jest.fn();
+
+  addCacheHeaders(fastify, null, done);
+
+  expect(fastify.addHook).toHaveBeenCalled();
+  expect(done).toHaveBeenCalled();
+  expect(reply.header).not.toHaveBeenCalled();
 });
--- a/__tests__/server/plugins/addCacheHeaders.spec.js
+++ b/__tests__/server/plugins/addCacheHeaders.spec.js
@@ -16,46 +16,48 @@
 
 import addCacheHeaders from '../../../src/server/plugins/addCacheHeaders';
 
-test('adds cache headers', () => {
-  const request = {
-    method: 'get',
-  };
-  const reply = {
-    header: jest.fn(),
-  };
-  const fastify = {
-    addHook: jest.fn(async (_hook, cb) => {
-      await cb(request, reply);
-    }),
-  };
-  const done = jest.fn();
+describe('addCacheHeaders', () => {
+  it('adds cache headers', () => {
+    const request = {
+      method: 'get',
+    };
+    const reply = {
+      header: jest.fn(),
+    };
+    const fastify = {
+      addHook: jest.fn(async (_hook, cb) => {
+        await cb(request, reply);
+      }),
+    };
+    const done = jest.fn();
 
-  addCacheHeaders(fastify, null, done);
+    addCacheHeaders(fastify, null, done);
 
-  expect(fastify.addHook).toHaveBeenCalled();
-  expect(done).toHaveBeenCalled();
-  expect(reply.header).toHaveBeenCalledTimes(2);
-  expect(reply.header).toHaveBeenCalledWith('Cache-Control', 'no-store');
-  expect(reply.header).toHaveBeenCalledWith('Pragma', 'no-cache');
-});
+    expect(fastify.addHook).toHaveBeenCalled();
+    expect(done).toHaveBeenCalled();
+    expect(reply.header).toHaveBeenCalledTimes(2);
+    expect(reply.header).toHaveBeenCalledWith('Cache-Control', 'no-store');
+    expect(reply.header).toHaveBeenCalledWith('Pragma', 'no-cache');
+  });
 
-test('does not add cache headers', () => {
-  const request = {
-    method: 'post',
-  };
-  const reply = {
-    header: jest.fn(),
-  };
-  const fastify = {
-    addHook: jest.fn(async (_hook, cb) => {
-      await cb(request, reply);
-    }),
-  };
-  const done = jest.fn();
+  it('does not add cache headers', () => {
+    const request = {
+      method: 'post',
+    };
+    const reply = {
+      header: jest.fn(),
+    };
+    const fastify = {
+      addHook: jest.fn(async (_hook, cb) => {
+        await cb(request, reply);
+      }),
+    };
+    const done = jest.fn();
 
-  addCacheHeaders(fastify, null, done);
+    addCacheHeaders(fastify, null, done);
 
-  expect(fastify.addHook).toHaveBeenCalled();
-  expect(done).toHaveBeenCalled();
-  expect(reply.header).not.toHaveBeenCalled();
+    expect(fastify.addHook).toHaveBeenCalled();
+    expect(done).toHaveBeenCalled();
+    expect(reply.header).not.toHaveBeenCalled();
+  });
 });
--- a/__tests__/server/plugins/addFrameOptionsHeader.spec.js
+++ b/__tests__/server/plugins/addFrameOptionsHeader.spec.js
@@ -14,47 +14,81 @@
  * permissions and limitations under the License.
  */
 
-import addFrameOptionsHeader from '../../../src/server/middleware/addFrameOptionsHeader';
+import addFrameOptionsHeader from '../../../src/server/plugins/addFrameOptionsHeader';
+import { updateCSP } from '../../../src/server/plugins/csp';
 
-jest.mock('../../../src/server/middleware/csp', () => ({
-  getCSP: () => ({
-    'frame-ancestors': ['valid.example.com'],
-  }),
-}));
+beforeEach(() => {
+  updateCSP('');
+});
 
-describe('addFrameOptionsHeader', () => {
-  let req;
-
-  const set = jest.fn((key, value) => value);
-  const res = { set };
-  const next = jest.fn();
-
-  beforeEach(() => {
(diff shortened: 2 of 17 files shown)


src/server/plugins/addFrameOptionsHeader.js
```diff
@@ -14,20 +14,34 @@
...
-  );
+    if (matchedDomain) {
+      reply.header('X-Frame-Options', `ALLOW-FROM ${referer}`);
+    }
+  });
 
-  if (matchedDomain) {
-    res.set('X-Frame-Options', `ALLOW-FROM ${referer}`);
-  }
+  done();
```
dogpatch626, line 41: why do we call done here instead of next as previously done?

dogpatch626 (approved):

src/server/plugins/addFrameOptionsHeader.js
giulianok: this is a fastify naming convention (see https://www.fastify.io/docs/latest/Reference/Plugins/), of course we can call it `next` if we want, but it's better to follow fastify's naming convention to keep everything more fastify-like and make it easier to read and associate.

you'll also see that i have renamed `req` -> `request` and `res` -> `reply`, same idea

dogpatch626: yea, I just thought that next and done where different in Fastify, good to know

giulianok: in this case, there isn't an equivalent in express since a plugin does more things than an middleware and a middleware gets executed on a request, but yeah they're similar

code-forger (approved):

__tests__/server/plugins/addSecurityHeaders.spec.js
```diff
@@ -14,39 +14,100 @@
...
+      await cb(request, reply);
+    }),
+  };
+  const done = jest.fn();
+
+  addSecurityHeaders(fastify, {}, done);
+
+  expect(fastify.addHook).toHaveBeenCalled();
+  expect(done).toHaveBeenCalled();
+  expect(reply.header).toHaveBeenCalledTimes(8);
```
bishnubista, line 41: Could have used foreach like on existing? 
    `expect(res.set.mock.calls.length).toEqual(Object.keys(securityHeaders).length);
    Object.keys(securityHeaders).forEach(...`

__tests__/server/plugins/ensureCorrelationId.spec.js
```diff
@@ -14,54 +14,52 @@
...
-    const { req, res, next } = generateSamples();
-    delete req.headers['correlation-id'];
-    ensureCorrelationId(req, res, next);
-    expect(req.headers).toHaveProperty('correlation-id');
-    expect(typeof req.headers['correlation-id']).toBe('string');
+    const { fastify, request } = generateSamples();
+    delete request.headers['correlation-id'];
+    ensureCorrelationId(fastify, null, jest.fn());
+    expect(request.headers).toHaveProperty('correlation-id');
+    expect(typeof request.headers['correlation-id']).toBe('string');
```
bishnubista, line 56: Does it add unique string?

__tests__/server/plugins/ensureCorrelationId.spec.js
```diff
@@ -14,54 +14,52 @@
...
-    expect(req.headers).toHaveProperty('correlation-id', 'thunderstorms');
+    const { fastify, request } = generateSamples();
+    delete request.headers['correlation-id'];
+    delete request.headers.correlation_id;
+    request.headers.unique_id = 'thunderstorms';
+    ensureCorrelationId(fastify, null, jest.fn());
+    expect(request.headers).toHaveProperty('correlation-id', 'thunderstorms');
   });
 
   it('adds a correlation-id header to a request without anything', () => {
```
bishnubista, line 51:
```suggestion
  it('adds a unique correlation-id to request header if correlation-id is null', () => {
```

src/server/plugins/addSecurityHeaders.js
```diff
@@ -14,11 +14,39 @@
...
+/**
+ * Fastify Plugin that adds security into headers
+ * @param {import('fastify').FastifyInstance} fastify Fastify instance
+ * @param {import('fastify').FastifyPluginOptions} _opts plugin options
+ * @param {import('fastify').FastifyPluginCallback} done plugin callback
+ */
+const addSecurityHeaders = (fastify, opts = {}, done) => {
+  const ignoreRoutes = opts.ignoreRoutes || [];
+
+  fastify.addHook('onRequest', async (request, reply) => {
```
bishnubista, line 28: Do you need async here?

src/server/plugins/csp.js
```diff
@@ -56,29 +57,43 @@ export function getCSP() {
...
+/**
+ * CSP Fastify Plugin
+ * @param {import('fastify').FastifyInstance} fastify Fastify instance
+ * @param {import('fastify').FastifyPluginOptions} _opts plugin options
+ * @param {import('fastify').FastifyPluginCallback} done plugin callback
+ */
+const csp = (fastify, _opts, done) => {
+  fastify.decorateRequest('scriptNonce', null);
+
+  fastify.addHook('onRequest', async (request, reply) => {
```
bishnubista, line 69: Same as above. Is `async` intentional?

__tests__/server/plugins/ensureCorrelationId.spec.js
giulianok: I propose `adds a unique correlation-id to the request object` since tests are assertions, things like `without` and `if` shouldn't be used. Also, these are the original test cases, maybe we should spend some time refreshing them 🤔

src/server/plugins/ensureCorrelationId.js
```diff
@@ -14,15 +14,28 @@
...
-  next();
-}
+/**
+ * Fastify Plugin that adds correlation id into headers
+ * @param {import('fastify').FastifyInstance} fastify Fastify instance
+ * @param {import('fastify').FastifyPluginOptions} _opts plugin options
+ * @param {import('fastify').FastifyPluginCallback} done plugin callback
+ */
+const ensureCorrelationId = (fastify, _opts, done) => {
+  fastify.addHook('onRequest', async (request) => {
```
bishnubista, line 27: Is `async` intentional?

src/server/plugins/forwardedHeaderParser.js
```diff
@@ -14,19 +14,33 @@
...
+  }), {});
+
+/**
+ * Fastify Plugin that injects 'forwarded' into the request object
+ * @param {import('fastify').FastifyInstance} fastify Fastify instance
+ * @param {import('fastify').FastifyPluginOptions} _opts plugin options
+ * @param {import('fastify').FastifyPluginCallback} done plugin callback
+ */
+const forwardedHeaderParser = (fastify, _opts, done) => {
+  fastify.addHook('onRequest', async (request) => {
```
bishnubista, line 34: Is async intentional?

src/server/plugins/setAppVersionHeader.js
```diff
@@ -14,11 +14,21 @@
...
 import readJsonFile from '../utils/readJsonFile';
 
 const { buildVersion: appVersion } = readJsonFile('../../../.build-meta.json');
 
-export default function setAppVersionHeader(_req, res, next) {
-  res.set('One-App-Version', appVersion);
-  next();
-}
+const setAppVersionHeader = (fastify, _opts, done) => {
+  fastify.addHook('onRequest', async (_request, reply) => {
```
bishnubista, line 24: Is async intentional?

__tests__/server/plugins/addSecurityHeaders.spec.js
giulianok: we could, what do you think is the advantage of the approach you're proposing?

bishnubista: Just cleaner approach in my view. But this works.

src/server/plugins/addSecurityHeaders.js
```diff
@@ -14,11 +14,39 @@
...
+    reply.header('Strict-Transport-Security', 'max-age=15552000; includeSubDomains');
+    reply.header('x-dns-prefetch-control', 'off');
```
10xLaCroixDrinker, line 30: Why does this include so many headers that were not there previously?

__tests__/server/plugins/csp.spec.js
```diff
@@ -15,170 +15,159 @@
...
+import Fastify from 'fastify';
+import csp, { updateCSP, getCSP, cspCache } from '../../../src/server/plugins/csp';
 
 const sanitizeCspString = (cspString) => cspString
-// replaces dynamic ip to prevent snapshot failures
+  // replaces dynamic ip to prevent snapshot failures
   // eslint-disable-next-line unicorn/better-regex -- conflicts with unsafe-regex
   .replace(/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}/g, '0.0.0.0');
 
-describe('csp', () => {
```
10xLaCroixDrinker, line 25: Please restore the describe block

__tests__/server/plugins/conditionallyAllowCors.spec.js
```diff
@@ -17,69 +17,92 @@
 // Dangling underscores are part of the HTTP mocks API
 /* eslint-disable no-underscore-dangle */
 import httpMocks from 'node-mocks-http';
+import fastifyCors from '@fastify/cors';
 import { fromJS } from 'immutable';
+import conditionallyAllowCors, { setCorsOrigins } from '../../../src/server/plugins/conditionallyAllowCors';
 
-describe('conditionallyAllowCors', () => {
```
10xLaCroixDrinker, line 22: Restore the describe block

__tests__/server/plugins/addSecurityHeaders.spec.js
```diff
@@ -14,39 +14,100 @@
  * permissions and limitations under the License.
  */
 
-import addSecurityHeaders from '../../../src/server/middleware/addSecurityHeaders';
-
-describe('addSecurityHeaders', () => {
```
10xLaCroixDrinker, line 19: Restore the describe block

__tests__/server/plugins/addFrameOptionsHeader.spec.js
```diff
@@ -14,47 +14,81 @@
...
 
-    expect(res.set).not.toHaveBeenCalled();
-    expect(next).toBeCalled();
+    expect(fastify.addHook).toHaveBeenCalled();
+    expect(done).toHaveBeenCalled();
+    expect(reply.header).not.toHaveBeenCalled();
   });
 });
+
+test('adds frame options header', () => {
```
10xLaCroixDrinker, line 71: Our convention is to use `it` over `test` and for all tests to be contained in a described block

__tests__/server/plugins/addCacheHeaders.spec.js
```diff
@@ -14,23 +14,48 @@
...
-import addCacheHeaders from '../../../src/server/middleware/addCacheHeaders';
+import addCacheHeaders from '../../../src/server/plugins/addCacheHeaders';
 
-describe('addCacheHeaders', () => {
-  it('should add all expected cache headers', () => {
-    const req = { get: jest.fn(), headers: {} };
-    const res = { set: jest.fn((key, value) => value) };
-    const next = jest.fn();
-    addCacheHeaders(req, res, next);
+test('adds cache headers', () => {
```
10xLaCroixDrinker, line 19: Use `it` and restore the describe block

10xLaCroixDrinker (changes requested):

src/server/plugins/addSecurityHeaders.js
giulianok: it is, fastify detects it and removes the need to call the `done` fn

src/server/plugins/csp.js
giulianok: yes

src/server/plugins/ensureCorrelationId.js
giulianok: yes

src/server/plugins/forwardedHeaderParser.js
giulianok: yes

src/server/plugins/setAppVersionHeader.js
giulianok: yes

src/server/plugins/addSecurityHeaders.js
giulianok: these are headers required by the integration tests that are not widely present. For the case of `x-dns-prefetch-control`, it is added by Helmet (see https://github.com/helmetjs/helmet/blob/a2549bd4688d9e711aea64a7fdffb07f4136f6cf/middlewares/x-dns-prefetch-control/index.ts#L17). Helmet, in the Fastify implementation, is only used on routes that will render Html, it isn't needed for all urls. By doing so, we increase performance. Right now, there's an integration test that expects this header in a route that doesn't render html (see https://github.com/americanexpress/one-app/blob/822999488c237a66298652d5746af65400234fd4/__tests__/integration/one-app.spec.js#L1365)

if we feel that we do not actually need them for all routes them we can delete them, they were added to support backwards compatibility

dogpatch626 (approved):

code-forger (approved):

bishnubista (approved):

10xLaCroixDrinker (approved):

PR #861, opened 2022-11-09
Feat/fastify holocron cdn

dogpatch626:
migrating one-app-holocron-dev-cdn into one app itself while refactoring it to use Fastify 
## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


__tests__/server/devHolocronCdn.spec.js
```diff
@@ -0,0 +1,58 @@
+/*
```
giulianok, line 1: curious why it went from `devHolocronCDN.spec.js` to `devHolocronCdn.spec.js`, I believe that that file name change makes github treat them as different files (one being deleted and another as new)

dogpatch626: Yeaaa see what happened is that I deleted it originally I believe thinking that it was not needed for coverage but then I re added it not undoing the deletion but re adding the file and I guess I messed up the name

src/server/utils/devCdnFactory.js
```diff
@@ -0,0 +1,188 @@
...
+  oneAppDevCdn.register(cors, {
+    origin: [
+      `http://localhost:${appPort}`,
+      `http://${ip.address()}:${appPort}`,
+      undefined,
+    ],
+  });
+
+  // for locally served modules
+  // oneAppDevCdn.use('/modules', express.static());
```
giulianok, line 70: remove

src/server/utils/devCdnFactory.js
```diff
@@ -0,0 +1,188 @@
...
+  oneAppDevCdn.get(`${routePrefix}/module-map.json`, async (req, reply) => {
+    const hostAddress = useHost ? `http://${req.headers.host}` : `http://localhost:${process.env.HTTP_ONE_APP_DEV_CDN_PORT}`;
+
+    const remoteMapRequest = remoteModuleMapUrl
+      ? await got(remoteModuleMapUrl, {
+        agent: {
+          https: new ProxyAgent(),
+          http: new ProxyAgent(),
+        },
+      }).then(
```
giulianok, line 90: by replacing the `then` with an `await` will make the code more readable

src/server/utils/devCdnFactory.js
```diff
@@ -0,0 +1,188 @@
...
+        (r) => {
+          // clear out remoteModuleBaseUrls as the new module map now has different urls in it
+          // not clearing would result in an ever growing array
+          remoteModuleBaseUrls = [];
+          const remoteModuleMap = JSON.parse(r.body);
+          const { modules } = remoteModuleMap;
+          const oneAppDevStaticsAddress = `${hostAddress}/static`;
+          Object.keys(modules).forEach((moduleName) => {
```
giulianok, line 98: `map` is a better fit than `forEach` since it will avoid mutation and generate a new array that can be assigned to `remoteModuleBaseUrls`

src/server/utils/devCdnFactory.js
```diff
@@ -0,0 +1,188 @@
...
+      })
+      : {};
+    const localMapRequest = useLocalModules ? JSON.parse(getLocalModuleMap({
+      pathToModuleMap: path.join(localDevPublicPath, 'module-map.json'),
+      oneAppDevCdnAddress: hostAddress,
+    })) : {};
+
+    // remoteMap always fulfilled
+    const remoteMap = remoteMapRequest;
+    const localMap = localMapRequest;
```
giulianok, line 135: are these assignments needed?

src/server/utils/devCdnFactory.js
```diff
@@ -0,0 +1,188 @@
...
+      const remoteModuleBaseUrlOrigin = new URL(knownRemoteModuleBaseUrl).origin;
+      got(`${remoteModuleBaseUrlOrigin}/${req.url}`, {
+
+        headers: { connection: 'keep-alive' },
+        agent: {
+          https: new ProxyAgent(),
+          http: new ProxyAgent(),
+        },
+      })
+        .then((remoteModuleResponse) => reply
```
giulianok, line 168: `await` would make the code more readable

giulianok (changes requested):

src/server/devHolocronCDN.js
```diff
@@ -18,36 +18,20 @@
 
 import path from 'path';
 import fs from 'fs';
-import express from 'express';
-import cors from 'cors';
 import { argv } from 'yargs';
-import Fastify from 'fastify';
-import fastifyExpress from '@fastify/express';
-import oneAppDevCdn from '@americanexpress/one-app-dev-cdn';
+import oneAppDevCdn from './utils/devCdnFactory';
```
giulianok, line 22: do we really need the business logic in `utils/devCdnFactory`? why can't we just put all the code from `utils/devCdnFactory` in this file?

giulianok (changes requested):

__tests__/server/devHolocronCdn.spec.js
giulianok: it would be great to maintain the same name since the tested file is actually called `devHolocronCDN` and to get a diff from github and see what have changed (easier to review)

PR #880, opened 2022-12-07
feat(fastifyholocroncdn): fix git commits

dogpatch626:
Migration and refactor of devHolocronCdn
## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


src/server/utils/devCdnFactory.js
```diff
@@ -0,0 +1,194 @@
...
+    const response = await got(remoteModuleMapUrl, {
+      agent: {
+        https: new ProxyAgent(),
+        http: new ProxyAgent(),
+      },
+    });
+
+    // clear out remoteModuleBaseUrls as the new module map now has different urls in it
+    // not clearing would result in an ever growing array
+    remoteModuleBaseUrls.splice(0, remoteModuleBaseUrls.length);
```
dogpatch626, line 43: as a preemptive explanation of this line, I had to clear the array in place as if I just re-assigned it like so 
```js remoteModuleBaseUrls = []``` it does not modify the original array which needs to be done

PR #882, opened 2022-12-07
feat(fastifycdn): migrate holocron cdn to fastify

dogpatch626:
holocron dev cdn migration
## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


__tests__/server/utils/devCdnFactory.spec.js
```diff
@@ -0,0 +1,674 @@
+/*
+ * Copyright 2019 American Express Travel Related Services Company, Inc.
```
code-forger, line 2:
```suggestion
 * Copyright 2022 American Express Travel Related Services Company, Inc.
```

src/server/utils/devCdnFactory.js
```diff
@@ -0,0 +1,194 @@
...
+) => remoteModuleBaseUrls.find((remoteModuleBaseUrl) => {
+  const remoteModuleUrlOrigin = new URL(remoteModuleBaseUrl).origin;
+  const remoteModulePath = remoteModuleBaseUrl.replace(remoteModuleUrlOrigin, '');
+
+  return incomingRequestPath.startsWith(remoteModulePath);
+});
+
+const consumeRemoteRequest = async (remoteModuleMapUrl, hostAddress, remoteModuleBaseUrls) => {
+  try {
+    const response = await got(remoteModuleMapUrl, {
```
code-forger, line 34: Why use `got` and not `node-fetch` (like the rest of our systems do)?

dogpatch626: Im honestly not entirely sure, It might be because of the use for proxyAgent Im not sure if node-fetch supports that natively?

dogpatch626: to do it apparently we need to use a package https-proxy-agent as long as proxy agent to make it work

Matthew-Mallimo: https://www.npmjs.com/package/node-fetch#custom-agent

bishnubista (approved):

package.json
```diff
@@ -110,6 +110,7 @@
     "express-rate-limit": "^6.5.1",
     "fastify": "^4.10.2",
     "fastify-plugin": "^4.2.0",
+    "got": "^11.8.3",
```
Matthew-Mallimo, line 113: We should remove this

src/server/utils/devCdnFactory.js
```diff
@@ -0,0 +1,199 @@
...
+
+  if (remoteModuleMapUrl) {
+    console.log(`one-app-dev-cdn loading module map from ${remoteModuleMapUrl}`);
+  } else {
+    console.log('one-app-dev-cdn only using locally served modules');
+  }
+
+  const oneAppDevCdn = Fastify();
+
+  if (process.env.NODE_ENV === 'production') {
```
giulianok, line 110: this check is already in place (see https://github.com/americanexpress/one-app/blob/feature/express-fastify-migration/src/server/index.js#L149). I don't think this is very useful since it returns an empty fastify instance, the server can be initialized but no routes will be available. I suggest to just throw an error, and even before creating the Fastify instance

src/server/utils/devCdnFactory.js
```diff
@@ -0,0 +1,199 @@
...
+
+    const localMap = useLocalModules ? JSON.parse(getLocalModuleMap({
+      pathToModuleMap: path.join(localDevPublicPath, 'module-map.json'),
+      oneAppDevCdnAddress: hostAddress,
+    })) : {};
+    let remoteMap = {};
+
+    if (remoteModuleMapUrl !== undefined) {
+      remoteMap = await consumeRemoteRequest(remoteModuleMapUrl, hostAddress, remoteModuleBaseUrls);
+    }
```
giulianok, line 148: we can simplify it by just converting it into an inline condition statement, e.g.

```
const remoteMap = remoteModuleMapUrl != null ? await consumeRemoteRequest(remoteModuleMapUrl, hostAddress, remoteModuleBaseUrls) : {}
```

Mutation makes code less readable because we have to search for the parts where the mutation happens, in this case, it can be avoided

giulianok (changes requested):

giulianok (approved):

PR #939, opened 2023-03-03
fix(load-modules): unnecessary calls to updateModuleRegistry

JAdshead:
## Description

currently any module which does not get loaded correctly is re-required on every module map poll.
This change limits fetching and parsing modules to one round per change in the module map.

## Motivation and Context

limit unnecessary calls to require-from-string which consumes memory.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

None

### DIFF
--- a/src/server/utils/loadModules.js
+++ b/src/server/utils/loadModules.js
@@ -26,17 +26,17 @@ import { setClientModuleMapCache } from './clientModuleMapCache';
 import { updateCSP } from '../plugins/csp';
 import addBaseUrlToModuleMap from './addBaseUrlToModuleMap';
 
-let cashedModuleMapHash;
+let cachedModuleMapHash;
 
 const loadModules = async () => {
   const moduleMapResponse = await fetch(process.env.HOLOCRON_MODULE_MAP_URL);
   const moduleMap = addBaseUrlToModuleMap(await moduleMapResponse.json());
 
   const moduleMapHash = hash(moduleMap);
-  if (cashedModuleMapHash && cashedModuleMapHash === moduleMapHash) {
+  if (cachedModuleMapHash && cachedModuleMapHash === moduleMapHash) {
     return {};
   }
-  cashedModuleMapHash = moduleMapHash;
+  cachedModuleMapHash = moduleMapHash;
   const serverConfig = getServerStateConfig();
 
   const loadedModules = await updateModuleRegistry({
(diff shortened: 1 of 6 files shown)


src/server/utils/loadModules.js
```diff
@@ -25,9 +26,17 @@ import { setClientModuleMapCache } from './clientModuleMapCache';
...
 
+let cashedModuleMapHash;
+
 const loadModules = async () => {
   const moduleMapResponse = await fetch(process.env.HOLOCRON_MODULE_MAP_URL);
   const moduleMap = addBaseUrlToModuleMap(await moduleMapResponse.json());
+
+  const moduleMapHash = hash(moduleMap);
+  if (cashedModuleMapHash && cashedModuleMapHash === moduleMapHash) {
+    return {};
```
code-forger, line 37: Is there value in logging here, maybe in the non-cache-hit case? so those who watch logs can see '3 modules change since last poll' and have confirmation that the change has actually been processed?

src/server/utils/loadModules.js
```diff
@@ -25,9 +26,17 @@ import { setClientModuleMapCache } from './clientModuleMapCache';
...
+  }
+  cashedModuleMapHash = moduleMapHash;
```
10xLaCroixDrinker, line 39:
```suggestion
let cachedModuleMapHash;

const loadModules = async () => {
  const moduleMapResponse = await fetch(process.env.HOLOCRON_MODULE_MAP_URL);
  const moduleMap = addBaseUrlToModuleMap(await moduleMapResponse.json());

  const moduleMapHash = hash(moduleMap);
  if (cachedModuleMapHash && cachedModuleMapHash === moduleMapHash) {
    return {};
  }
  cascedModuleMapHash = moduleMapHash;
```

Minor spelling issue

src/server/utils/loadModules.js
10xLaCroixDrinker: We already log that

https://github.com/americanexpress/one-app/blob/531958f5d20b30900efd34db88fae7a6c5e8406b/src/server/utils/pollModuleMap.js#L151

10xLaCroixDrinker (changes requested):

10xLaCroixDrinker (approved):

Matthew-Mallimo (approved):

PR #963, opened 2023-03-29
feat(server): messages of missing local modules in devHolocronCDN

PixnBits:
## Description

use a @americanexpress/one-app-dev-cdn hook `localModuleBundleNotFound` to log messages that an expected/required module bundle is missing and what the user should do to resolve the issue

pairs with https://github.com/americanexpress/one-app-dev-cdn/pull/47

## Motivation and Context

A scenario observed to be common is a user serving multiple modules locally with @americanexpress/one-app-runner but forgetting to build one (different directory). Determining why the module fails to load and what to do about it/their course of action can be tricky, requiring deciphering the one-app log messages .

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

The intention is to aid Developers by reducing their time to diagnosis and resolution by more explicitly identifying a situation and the usual action(s) they need to take.

### DIFF
--- a/src/server/devHolocronCDN.js
+++ b/src/server/devHolocronCDN.js
@@ -22,8 +22,41 @@ import express from 'express';
 import cors from 'cors';
 import { argv } from 'yargs';
 import oneAppDevCdn from '@americanexpress/one-app-dev-cdn';
+import { register } from 'prom-client';
+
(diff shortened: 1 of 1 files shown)


PixnBits:
I'm thinking there's probably a better approach

PR #966, opened 2023-03-31
WIP - enable fallback externals

JAdshead:
## Description

This is rough early ideas for handling fallback externals

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


src/server/utils/onModuleLoad.js
```diff
@@ -150,21 +157,19 @@ export default function onModuleLoad({
...
         messages.push(`External '${externalName}' is required by ${moduleName}, but is not provided by the root module`);
+        registerExternalFallback({ moduleName, externalName, requestedExternalVersion });
       } else if (!semver.satisfies(providedExternal.version, requestedExternalVersion)) {
         const failedExternalMessage = `${externalName}@${requestedExternalVersion} is required by ${moduleName}, but the root module provides ${providedExternal.version}`;
-        if (process.env.ONE_DANGEROUSLY_ACCEPT_BREAKING_EXTERNALS) {
-          console.warn(failedExternalMessage);
-        } else {
-          messages.push(failedExternalMessage);
-        }
+        registerExternalFallback({ moduleName, externalName, requestedExternalVersion });
```
giulianok, line 166: I guess the module will call a function for that particular external (e.g. `module.exports = something()` that will try to find the module's externals in the Map, if it isn't available then go to the Root Module and look for it, and, if can't still find it them reject the module

PR #982, opened 2023-04-24
chore(update-dev-proxy): update dev proxy to 2 replace require with i…

dogpatch626:
Update dev proxy to v2. one-app-dev-proxy received updates that required it to be imported rather than required.

## Types of Changes

- [x] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/src/server/index.js
+++ b/src/server/index.js
@@ -22,13 +22,14 @@ import enData from 'lean-intl/locale-data/json/en.json';
 
 import './init';
 // Allow env config to run before importing holocron. (x2)
+// eslint-disable-next-line import/no-extraneous-dependencies
+import oneAppDevProxy from '@americanexpress/one-app-dev-proxy';
 import ssrServer from './ssrServer';
 import metricsServer from './metricsServer';
 import { addServer, shutdown } from './shutdown';
 import pollModuleMap from './utils/pollModuleMap';
 import loadModules from './utils/loadModules';
 import getHttpsConfig from './utils/getHttpsConfig';
-
 /*    🛠️ UTILITIES 🛠️    */
 
 const getRemotesFromDevEndpointsFile = () => {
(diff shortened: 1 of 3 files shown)


JAdshead (approved):

src/server/index.js
```diff
@@ -115,9 +114,9 @@ async function devHolocronCDNStart() {
 
 async function oneAppDevProxyStart() {
   const { argv } = require('yargs');
-  // eslint-disable-next-line import/no-extraneous-dependencies
-  const oneAppDevProxy = require('@americanexpress/one-app-dev-proxy');
   const oneAppDevProxyPort = process.env.HTTP_ONE_APP_DEV_PROXY_SERVER_PORT;
+  // eslint-disable-next-line import/no-extraneous-dependencies
+  const oneAppDevProxy = require('@americanexpress/one-app-dev-proxy').default;
```
giulianok, line 119: why don't we use modules (import) instead?

dogpatch626: This is how We do it for other packages that are cjs, @JAdshead  and I talked through it a bit and just decided to go this way

JAdshead: This needs to be an import statement as `import` can't be nested inside functions. This should only be required when the server is running in development hence the need for `require`. 

In regards to needing the `.default`  [one-app-dev-proxy](https://github.com/americanexpress/one-app-dev-proxy/blob/main/src/index.js) uses ES module syntax `export default server` which gets transpiled to `module.exports = { default: server }`. 

Webpack does some magic for `import` statements and will transpile them to `const dep = require('x').default`  when whats being imported is already a webpack tranpiled es module `requiredModule.__esModule: true`. 

The reason for now needed the `.default` is with the last release of `one-app-dev-proxy`, it updated the babel version. Babel used to try and have some extra magic where if a ES module only had a single default export, ie no named exports, it would transpile to `module.exports = server` while also patching the `module` object to include `default` 

Cleaner explanation can been seen in [this stackoverflow](https://stackoverflow.com/questions/33704714/cant-require-default-export-value-in-babel-6-x)

JAdshead (approved):

Matthew-Mallimo (approved):

PR #984, opened 2023-04-25
feat(external-fallbacks): enable modules to have external fallbacks

JAdshead:
This PR requires release and update of Holocron with the changes outlined in https://github.com/americanexpress/holocron/pull/143

## Description

Enable modules which require externals to use a fallback when the root module does not provide one.
Record each modules required externals when the module is required by one-app.
When rendering html to the client include all the fallback script tags required for the modules used in the initial render.
pass the registry of required externals down to the client then hydrate when client initializes 

## Motivation and Context

This will allow more flexibility to holocron modules when a required external is missing. It also provides an additional upgrade path for major versions of external dependencies. 

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


src/server/utils/onModuleLoad.js
```diff
@@ -32,25 +33,12 @@ import { setErrorPage } from '../plugins/reactHtml/staticErrorPage';
...
-
-export const getModulesUsingExternals = () => modulesUsingExternals.toJS();
 
-export const setModulesUsingExternals = (moduleNames) => {
-  modulesUsingExternals = new ImmutableSet(moduleNames);
-};
+export const CONFIGURATION_KEY = 'appConfig';
 
-const logModuleLoad = (moduleName, moduleVersion) => {
+function logModuleLoad(moduleName, moduleVersion) {
```
Matthew-Mallimo, line 39: why was this changed?

src/server/utils/onModuleLoad.js
```diff
@@ -147,32 +133,6 @@ export default function onModuleLoad({
     );
   }
 
-  if (requiredExternals) {
```
Matthew-Mallimo, line 150: Did all this logic get moved to holocron?

package.json
```diff
@@ -108,7 +108,7 @@
     "fastify": "^4.10.2",
     "fastify-plugin": "^4.2.0",
     "helmet": "^6.0.0",
-    "holocron": "^1.7.0",
+    "holocron": "file:holocron-v1.7.0.tgz",
```
Matthew-Mallimo, line 111: This will need to be fixed before merging

Matthew-Mallimo (changes requested):

src/client/initClient.jsx
```diff
@@ -29,6 +29,8 @@ export default async function initClient() {
   try {
     // eslint-disable-next-line no-underscore-dangle
     setModuleMap(global.__CLIENT_HOLOCRON_MODULE_MAP__);
+    // eslint-disable-next-line no-underscore-dangle
+    setRequiredExternalsRegistry(global.__holocron_externals__);
```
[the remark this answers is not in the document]
giulianok: good point, there are still other lower case global vars but I like the consistency of using all caps for it

src/server/plugins/reactHtml/index.jsx
```diff
@@ -87,12 +99,45 @@ export function renderModuleScripts({
...
+      const { baseUrl } = modules[moduleName];
+      const endsWithSlash = baseUrl.endsWith('/');
+      const src = `${baseUrl}${endsWithSlash ? '' : '/'}${name}.browser.js`;
+      return renderScript({
+        src,
+        integrity,
+        isDevelopmentEnv,
+        clientCacheRevision,
+      });
+    }).join(isDevelopmentEnv ? '\n          ' : '');
```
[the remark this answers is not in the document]
giulianok: idk to be honest. @JAdshead any reason why you've added these spaces?

src/server/plugins/reactHtml/index.jsx
```diff
@@ -87,12 +99,45 @@ export function renderModuleScripts({
...
+    return renderScript({
+      src, integrity, isDevelopmentEnv, clientCacheRevision,
+    });
   }).join(isDevelopmentEnv ? '\n          ' : '');
 }
 
+function isDuplicateExternal({ name, version }, currentExternals) {
+  return currentExternals
+    .some((current) => (name === current.name) && (version === current.version));
+}
```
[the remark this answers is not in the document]
giulianok: if you are referring to the `filter` + `isDuplicateExternal` then, technically, it's `O(nm)`, and yes, can be `O(n^2)` depending on the data it each array holds. Also, both arrays are small, no big impact, but I'm all for performance so happy to convert it into a set for quick lookup

src/server/utils/onModuleLoad.js
giulianok: consistency

src/server/utils/onModuleLoad.js
giulianok: yes

__tests__/server/utils/getModulesToUpdate.spec.js
```diff
@@ -152,7 +152,7 @@ describe('getModulesToUpdate', () => {
         },
       },
     };
-    expect(getModulesToUpdate(current, next)).toEqual(['first-module', 'another-module', 'test-root']);
+    expect(getModulesToUpdate(current, next)).toEqual(['test-root']);
```
Matthew-Mallimo, line 155: Why did this expect change if the test case didn't change?

package.json
```diff
@@ -31,7 +31,7 @@
     "test:lockfile": "lockfile-lint -p package-lock.json -t npm -a npm -o https: -c -i",
     "test:lint": "eslint --ext js,jsx,md,snap .",
     "start": "node lib/server/index.js",
-    "start:inspect": "node --inspect --expose-gc lib/server/index.js",
+    "start:inspect": "node --inspect-brk --expose-gc lib/server/index.js",
```
Matthew-Mallimo, line 34:
```suggestion
    "start:inspect": "node --inspect --expose-gc lib/server/index.js",
```

Matthew-Mallimo (changes requested):

__tests__/server/utils/getModulesToUpdate.spec.js
JAdshead: good catch, looks like the mock outdated

package.json
JAdshead: resolved in 5072f48bf0080408e9584a2277514d411e3298ef

__tests__/server/utils/getModulesToUpdate.spec.js
JAdshead: resolved 5072f48bf0080408e9584a2277514d411e3298ef

Matthew-Mallimo (approved):

Matthew-Mallimo (approved):

Matthew-Mallimo (dismissed):

src/client/initClient.jsx
```diff
@@ -29,6 +29,8 @@ export default async function initClient() {
   try {
     // eslint-disable-next-line no-underscore-dangle
     setModuleMap(global.__CLIENT_HOLOCRON_MODULE_MAP__);
+    // eslint-disable-next-line no-underscore-dangle
+    setRequiredExternalsRegistry(global.__HOLOCRON_EXTERNALS__);
```
code-forger, line 33: NIT: for consistency with `__CLIENT_HOLOCRON_MODULE_MAP__`, why not `__CLIENT_HOLOCRON_EXTERNALS__`?

src/server/utils/onModuleLoad.js
```diff
@@ -33,21 +33,8 @@ import { setRedirectAllowList } from './redirectAllowList';
...
-
-const clearModulesUsingExternals = () => {
-  modulesUsingExternals = modulesUsingExternals.clear();
-};
 
-export const getModulesUsingExternals = () => modulesUsingExternals.toJS();
-
-export const setModulesUsingExternals = (moduleNames) => {
-  modulesUsingExternals = new ImmutableSet(moduleNames);
-};
```
code-forger, line 50: Removing parts of the old system? is this still backwards compatible and this functionality is now achieved in a different way?

code-forger (approved):
Approved with 1 NIT and 1 question

src/server/utils/onModuleLoad.js
JAdshead: for the most part this has been reimplemented in holocron - https://github.com/americanexpress/holocron/pull/143/files#diff-c59d732ae8137c4f28a025d9cabd09b4fd3166326f1822eb6b9a94089d89d777R68-R77

it does appear that the call to `clearModulesUsingExternals` has been removed when it should not have.

src/client/initClient.jsx
JAdshead: `__CLIENT_HOLOCRON_MODULE_MAP__` is the only key with the `CLIENT` prefix, this is probably as the module map being set is specifically for the client and differs from the module map loaded on the server.

Matthew-Mallimo (approved):

src/server/plugins/reactHtml/index.jsx
```diff
@@ -88,12 +100,60 @@ export function renderModuleScripts({
...
 }
 
+export function renderExternalFallbacks({
+  clientInitialState,
+  moduleMap,
+  isDevelopmentEnv,
+  isLegacy,
+}) {
+  const loadedModules = clientInitialState.getIn(['holocron', 'loaded'], iSet()).toArray();
+  const requiredFallbacks = loadedModules
```
dogpatch626, line 116: this is a pretty expensive looking reduce in terms of time. are we bound to this by the shape of the data that is in loadedModules or is there potentially a better way of doing this. nested objects with arrays are fun 😄

JAdshead: This would only be a handful of modules. loadedModules are modules which are required to render the page, not all the modules included in the module map. even them i'm not sure a module map would be ever be large enough to provide performance concerns. 

> are we bound to this by the shape of the data that is in loadedModules
yes, this could be rewritten to remove that additional reduce on 118 but im not sure that it would add much value.

dogpatch626 (approved):

Matthew-Mallimo (approved):

PR #992, opened 2023-05-09
fix(memoryLeak): disable cache for transitjs

Matthew-Mallimo:
## Description

We noticed a memory leak happening where if non-immutable data was put into the store, transit-js would cache all the keys for that non-immutable data. This cache persisted across calls, and would grow indefinitely. 

The solution is to use transit-js directly, and disable the cache. unfortunately transit-immutable-js does not allow you to pass options to transit-js, so we have to add handles manually.

## Motivation and Context

memory leaks are bad

## How Has This Been Tested?

Tested independently and unit tested

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/src/universal/utils/transit.js
+++ b/src/universal/utils/transit.js
@@ -18,7 +18,8 @@
 // the browser or on the server
 /* eslint-disable global-require */
 
-import transit from 'transit-immutable-js';
+import instance, { handlers } from 'transit-immutable-js';
+import transit from 'transit-js';
 import { serializeError } from 'serialize-error';
 
 const concealOrigin = (href) => href && href.replace(/\/\/[^/]+/g, '//***');
@@ -27,7 +28,6 @@ export function writeError(value) {
   const error = serializeError(value);
   delete error.stack;
   error.message = concealOrigin(error.message);
-
   if (error.response) {
     error.response.url = concealOrigin(error.response.url);
   }
@@ -35,7 +35,7 @@ export function writeError(value) {
   return error;
 }
 
-export default transit.withExtraHandlers([
+const extraHandlers = [
   {
     tag: 'error',
     class: Error,
@@ -57,4 +57,32 @@ export default transit.withExtraHandlers([
       : require('url').parse(value)
     ),
   },
-]);
+];
+
+const modifiedHandlers = handlers.withExtraHandlers(extraHandlers);
+
+const reader = transit.reader('json', {
+  handlers: modifiedHandlers.read,
+  cache: false,
+  // Copied from transit-immutable-js. Without this, Error fails to read.
+  mapBuilder: {
+    init() {
+      return {};
+    },
+    add(m, k, v) {
+      // eslint-disable-next-line no-param-reassign
+      m[k] = v;
+      return m;
+    },
+    finalize(m) {
+      return m;
+    },
+  },
+});
+const writer = transit.writer('json', { handlers: modifiedHandlers.write, cache: false });
+
+export default {
+  ...instance.withExtraHandlers(extraHandlers), // added for test cases
+  toJSON: (data) => writer.write(data),
+  fromJSON: (json) => reader.read(json),
+};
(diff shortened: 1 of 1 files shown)


PixnBits:
> ## Motivation and Context
> memory leaks are bad

😂

src/universal/utils/transit.js
```diff
@@ -57,4 +57,32 @@ export default transit.withExtraHandlers([
...
+    finalize(m) {
+      return m;
+    },
+  },
+});
+const writer = transit.writer('json', { handlers: modifiedHandlers.write, cache: false });
+
+export default {
+  ...instance.withExtraHandlers(extraHandlers), // added for test cases
+  toJSON: (data) => writer.write(data),
```
PixnBits, line 86: do these exports inflate the client bundle?
that is, can we separate the persistent server writer from the single-use (?) client reader? 🤔

src/universal/utils/transit.js
```diff
@@ -57,4 +57,32 @@ export default transit.withExtraHandlers([
...
+const reader = transit.reader('json', {
+  handlers: modifiedHandlers.read,
+  cache: false,
+  // Copied from transit-immutable-js. Without this, Error fails to read.
+  mapBuilder: {
+    init() {
+      return {};
+    },
+    add(m, k, v) {
+      // eslint-disable-next-line no-param-reassign
```
PixnBits, line 73:
```suggestion
      // eslint-disable-next-line no-param-reassign -- reason so aid updating eslint-config-amex 🙏 
```

src/universal/utils/transit.js
```diff
@@ -57,4 +57,32 @@ export default transit.withExtraHandlers([
...
   },
-]);
+];
+
+const modifiedHandlers = handlers.withExtraHandlers(extraHandlers);
+
+const reader = transit.reader('json', {
+  handlers: modifiedHandlers.read,
+  cache: false,
+  // Copied from transit-immutable-js. Without this, Error fails to read.
```
PixnBits, line 67: note: license is MIT
* https://github.com/glenjamin/transit-immutable-js/blob/v0.8.0/package.json#L18
* https://github.com/glenjamin/transit-immutable-js/blob/v0.8.0/LICENSE

src/universal/utils/transit.js
```diff
@@ -57,4 +57,32 @@ export default transit.withExtraHandlers([
...
+      return m;
+    },
+  },
+});
+const writer = transit.writer('json', { handlers: modifiedHandlers.write, cache: false });
+
+export default {
+  ...instance.withExtraHandlers(extraHandlers), // added for test cases
+  toJSON: (data) => writer.write(data),
+  fromJSON: (json) => reader.read(json),
```
PixnBits, line 87: can we delete the `reader` value (mark for garbage collection) for browser memory saving? 🤔 
maybe a follow up PR, if so

src/universal/utils/transit.js
```diff
@@ -57,4 +57,32 @@ export default transit.withExtraHandlers([
...
+    },
+    finalize(m) {
+      return m;
+    },
+  },
+});
+const writer = transit.writer('json', { handlers: modifiedHandlers.write, cache: false });
+
+export default {
+  ...instance.withExtraHandlers(extraHandlers), // added for test cases
```
PixnBits, line 85: like https://github.com/americanexpress/one-app/blob/main/__tests__/universal/utils/transit.spec.js#L66-L72
huh, are the other methods used? 🤔

PixnBits (changes requested):
IMO eslint exception override explanation required
others not, but nice to haves

src/universal/utils/transit.js
Matthew-Mallimo: I don't want to do that type of optimization in this PR. We are exporting almost exactly what was exported previously, so it may be a few bytes bigger, but I dont see that as an issue.

src/universal/utils/transit.js
Matthew-Mallimo: Follow-up, this keeps parity with whats already happening

code-forger (approved):

PR #993, opened 2023-05-09
fix(memoryLeak): disable cache for transitjs

Matthew-Mallimo:
## Description

We noticed a memory leak happening where if non-immutable data was put into the store, transit-js would cache all the keys for that non-immutable data. This cache persisted across calls, and would grow indefinitely. 

The solution is to use transit-js directly, and disable the cache. unfortunately transit-immutable-js does not allow you to pass options to transit-js, so we have to add handles manually.

## Motivation and Context

memory leaks are bad

## How Has This Been Tested?

Tested independently and unit tested

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/src/universal/utils/transit.js
+++ b/src/universal/utils/transit.js
@@ -18,7 +18,8 @@
 // the browser or on the server
 /* eslint-disable global-require */
 
-import transit from 'transit-immutable-js';
+import instance, { handlers } from 'transit-immutable-js';
+import transit from 'transit-js';
 import { serializeError } from 'serialize-error';
 
 const concealOrigin = (href) => href && href.replace(/\/\/[^/]+/g, '//***');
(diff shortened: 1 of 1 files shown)


src/universal/utils/transit.js
```diff
@@ -57,4 +57,32 @@ export default transit.withExtraHandlers([
...
   },
-]);
+];
+
+const modifiedHandlers = handlers.withExtraHandlers(extraHandlers);
+
+const reader = transit.reader('json', {
+  handlers: modifiedHandlers.read,
+  cache: false,
+  // Copied from transit-immutable-js. Without this, Error fails to read.
```
code-forger, line 67: can you reference a file, line, and commit hash for where you took this from.

That way when these deps are major'd, we can re-reference

JAdshead (approved):

code-forger (approved):

PixnBits (approved):

PR #994, opened 2023-05-09
chore(workflows): on-pr-and-trunk_one-app-unit-and-lint-tests exclude 18.16.0

PixnBits:
## Description

test with Node.js 18.x but before 18.16.0
This will need to be reverted when the bug is resolved and released in a new version.

## Motivation and Context
unit tests are affected by a bug: https://github.com/nodejs/node/issues/47563

## How Has This Been Tested?

GitHub workflow, so can't run locally :S

but did run the semver syntax through https://semver.npmjs.com/ with lodash via `^4 < 4.17`

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

PRs should have less false-positives.

### DIFF
--- a/.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml
+++ b/.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16.x, 18.x]
+        node-version: [16.x, "^18 < 18.16"]
     steps:
       - name: Checkout
         uses: actions/checkout@v3


.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16.x, 18.x]
+        node-version: [16.x, "^18 < 18.16"]
```
bishnubista, line 14: Great! I was looking into creating this PR.

JAdshead (approved):
I have updated branch permissions for this change. that will also need to be reverted when fix is released

Matthew-Mallimo (approved):

PR #1014, opened 2023-05-31
a server tracing system for One App

code-forger:
## Description
This system produces json logs with timing for how long each server phase, and all 'fetches' started, took during a single request to the server

## Motivation and Context
This allows visibility over what your server is doing, including finding sequences in API calls, and other bottle necks in your system.

## How Has This Been Tested?
By loading it into both non-production and staging-production websites, and by thorough unit testing

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

N/A


Matthew-Mallimo (approved):

src/server/middleware/createRequestHtmlFragment.jsx
dogpatch626, line 1: So is the Tracer allways technically on? what's opt in is to get the tracer logs?

src/server/config/env/runTime.js
```diff
@@ -237,6 +237,21 @@ const runTime = [
...
+  {
+    name: 'ONE_ENABLE_SERVER_TRACING',
+    defaultValue: 'false',
+    normalize: (input) => {
+      if (input.toLowerCase() === 'true') {
+        return 'true';
+      }
+      return 'false';
+    },
+    validate: (input) => {
```
giulianok, line 249: does `normalize` run before `validate`?

src/server/config/env/runTime.js
```diff
@@ -237,6 +237,21 @@ const runTime = [
       }
     },
   },
+  {
+    name: 'ONE_ENABLE_SERVER_TRACING',
+    defaultValue: 'false',
+    normalize: (input) => {
+      if (input.toLowerCase() === 'true') {
```
giulianok, line 244: looks like we have duplicate code for this, it would be nice to abstract it into a function to reuse the code, something like `...normalizeAndValidateBoolean()`

src/server/utils/tracer.js
```diff
@@ -0,0 +1,198 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+// The NoOpTracer is a close variant to the Tracer and so is ok in this file
+// eslint-disable-next-line max-classes-per-file
+export class Tracer {
```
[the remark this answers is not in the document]
code-forger: Object Orientation here offers an explicit advantage, Encapsulation and Abstraction. 

By encapsulating all the tracing functions into the Tracer class, we can (very syntactically cleanly) also create a NoOpTracer.

This allows the server to 'install' either Tracer or NoOpTracer, and all the other code in the system does not know the difference.

That way the check to ONE_ENABLE_SERVER_TRACER can be singular and central.

This also allows for the clean use of private fields and functions, and there is no ambiguity over the scope of all memory allocated by this system, as it can be garbage collected all at once.

src/server/middleware/createRequestHtmlFragment.jsx
code-forger: https://github.com/americanexpress/one-app/pull/1014/files/8b1395f77f11a2e23c10b385ac80cb4a880bb642#diff-c72b25996c9778fefcf24f888b3e2944a5948b5e0ec82320573ce0c97229b041R164

No, if the tracer is disabled, the server installs the NoOpTracer.

the tracer functions are always called, but they are no ops if the tracer is disabled, so it has neglegable memory or performance impact

src/server/config/env/runTime.js
code-forger: Good spot, since Normalize always returns a valid value, I've removed 'validate'

src/server/config/env/runTime.js
```diff
@@ -237,6 +237,21 @@ const runTime = [
       }
     },
   },
+  {
+    name: 'ONE_ENABLE_SERVER_TRACING',
+    defaultValue: 'false',
+    normalize: (input) => {
```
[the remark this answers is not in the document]
code-forger: Done

src/server/config/env/runTime.js
code-forger: This function was removed anyway

Matthew-Mallimo (approved):

JAdshead (approved):

src/server/utils/tracer.js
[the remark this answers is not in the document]
giulianok: encapsulation and easy to access to fns make sense in this context, I'm just always get a bit worried about contributors getting too passionate about classes and start extending them, making solutions unreadable and hard to maintain

JAdshead (dismissed):

Matthew-Mallimo (approved):

giulianok (approved):

PR #1024, opened 2023-06-15
feat(circuitBreaker): configurable event loop delay percentile

10xLaCroixDrinker:
## Description

New configuration option for circuit breakers event loop delay check

## Motivation and Context

Currently the circuit breaker is based on max event loop delay, which is not ideal as it indicates an outlier rather than a trend. p(99) can be significantly lower than p(100) (max). The default is left at 100 to avoid a breaking change to behavior. This change required the addition of an interval upon which to reset the histogram since it biases lower over time.

## How Has This Been Tested?

Unit test + running locally while monitoring circuit breaker metrics and load testing with autocanon

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

- [x] These changes should be applied to ~~a maintenance~~ **main** branch.

### DIFF
--- a/src/server/utils/createCircuitBreaker.js
+++ b/src/server/utils/createCircuitBreaker.js
@@ -22,12 +22,24 @@ import { registerCircuitBreaker } from '../metrics/circuit-breaker';
 
 const { rootModuleName } = getServerStateConfig();
 let eventLoopDelayThreshold = 250;
+let eventLoopDelayPercentile = 100;
 
 export const setEventLoopDelayThreshold = (n) => {
-  eventLoopDelayThreshold = Number(n) || 250;
+  eventLoopDelayThreshold = Number.parseInt(n, 10) || 250;
+};
+
+export const setEventLoopDelayPercentile = (n) => {
+  if (typeof n !== 'undefined' && (n !== Number.parseInt(n, 10) || n < 1 || n > 100)) {
+    console.warn(`Event loop percentile must be an integer in range 1-100; given ${JSON.stringify(n)}. Defaulting to p(100).`);
+    eventLoopDelayPercentile = 100;
+    return;
+  }
+  // Default to p(100) to avoid breaking change for users expecting max delay
+  eventLoopDelayPercentile = n || 100;
 };
 
 export const getEventLoopDelayThreshold = () => eventLoopDelayThreshold;
+export const getEventLoopDelayPercentile = () => eventLoopDelayPercentile;
 
 const options = {
   // Do not use a timeout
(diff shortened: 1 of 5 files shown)


src/server/utils/createCircuitBreaker.js
```diff
@@ -22,12 +22,24 @@ import { registerCircuitBreaker } from '../metrics/circuit-breaker';
 
 const { rootModuleName } = getServerStateConfig();
 let eventLoopDelayThreshold = 250;
+let eventLoopDelayPercentile = 100;
 
 export const setEventLoopDelayThreshold = (n) => {
```
JAdshead, line 27: This will always be called by root module even if not set resulting in n=undefined
```suggestion
export const setEventLoopDelayThreshold = (n=eventLoopDelayPercentile) => {
```

src/server/utils/onModuleLoad.js
```diff
@@ -129,6 +130,7 @@ export default function onModuleLoad({
     setConfigureRequestLog(configureRequestLog);
     setCreateSsrFetch(createSsrFetch);
     setEventLoopDelayThreshold(eventLoopDelayThreshold);
+    setEventLoopDelayPercentile(eventLoopDelayPercentile);
```
JAdshead, line 133:
```suggestion
    if(eventLoopDelayPercentile) { setEventLoopDelayPercentile(eventLoopDelayPercentile); }
```

src/server/utils/createCircuitBreaker.js
JAdshead, line 27: maybe set to default when no value given
```suggestion
export const setEventLoopDelayThreshold = (n=eventLoopDelayThreshold) => {
```

src/server/utils/createCircuitBreaker.js
```diff
@@ -41,16 +53,27 @@ const options = {
...
+  histogramResetInterval = setInterval(() => eventLoopDelayHistogram.reset(), 30e3);
+};
+
+const checkEventLoopDelay = async () => {
   if (process.env.NODE_ENV === 'development') return;
   // Return if root module is not loaded, as that is where threshold is configured
   if (!getModule(rootModuleName)) return;
   // Get event loop delay in milliseconds (from nanoseconds)
-  const maxEventLoopDelay = eventLoopDelayHistogram.max / 1e6;
+  const eventLoopDelay = eventLoopDelayHistogram.percentile(eventLoopDelayPercentile) / 1e6;
```
JAdshead, line 70: can this handle any number ?

10xLaCroixDrinker: Just integers 1-100, as validated

src/server/utils/createCircuitBreaker.js
```diff
@@ -22,12 +22,24 @@ import { registerCircuitBreaker } from '../metrics/circuit-breaker';
...
 let eventLoopDelayThreshold = 250;
+let eventLoopDelayPercentile = 100;
 
 export const setEventLoopDelayThreshold = (n) => {
-  eventLoopDelayThreshold = Number(n) || 250;
+  eventLoopDelayThreshold = Number.parseInt(n, 10) || 250;
+};
+
+export const setEventLoopDelayPercentile = (n) => {
+  if (typeof n !== 'undefined' && (n !== Number.parseInt(n, 10) || n < 1 || n > 100)) {
```
dogpatch626, line 32: why not just 
```js
  if (!n && (n !== Number.parseInt(n, 10) || n < 1 || n > 100)) {

```

JAdshead (approved):

PR #1029, opened 2023-06-16
fix(configureRequestLog): use default when configured with undefined

JAdshead:
## Description

if configureRequestLog is not provided by root module, request logs are not being logged.

## Motivation and Context

re enable request logs
## How Has This Been Tested?

Test suite and locally
## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/src/server/utils/logging/fastifyPlugin.js
+++ b/src/server/utils/logging/fastifyPlugin.js
@@ -201,12 +201,17 @@ const fastifyPlugin = (fastify, _opts, done) => {
   });
 
   fastify.addHook('onResponse', async (request, reply) => {
-    endTimer(request, $ResponseBuilder);
-    // same as 'reply.getResponseTime()' but our approach
-    // helps us to make the code cleaner
-    endTimer(request, $RequestFullDuration);
+    try {
+      endTimer(request, $ResponseBuilder);
+      // same as 'reply.getResponseTime()' but our approach
+      // helps us to make the code cleaner
+      endTimer(request, $RequestFullDuration);
 
-    logClientRequest(request, reply);
+      logClientRequest(request, reply);
+    } catch (error) {
+      console.error(error);
+      throw error;
+    }
   });
 
   done();
--- a/src/server/utils/logging/fastifyPlugin.js
+++ b/src/server/utils/logging/fastifyPlugin.js
@@ -35,8 +35,9 @@ export const $RequestOverhead = Symbol('$RequestOverhead');
 export const $RouteHandler = Symbol('$RouteHandler');
 export const $ResponseBuilder = Symbol('$ResponseBuilder');
 
+const passThrough = ({ log }) => log;
 const UTILS = {
-  configureRequestLog: ({ log }) => log,
+  configureRequestLog: passThrough,
 };
 
 const getLocale = (req) => {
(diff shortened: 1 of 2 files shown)


Matthew-Mallimo (approved):

src/server/utils/logging/fastifyPlugin.js
```diff
@@ -200,12 +201,17 @@ const fastifyPlugin = (fastify, _opts, done) => {
...
-    logClientRequest(request, reply);
+    try {
+      endTimer(request, $ResponseBuilder);
+      // same as 'reply.getResponseTime()' but our approach
+      // helps us to make the code cleaner
+      endTimer(request, $RequestFullDuration);
+
+      logClientRequest(request, reply);
+    } catch (error) {
+      console.error(error);
```
giulianok, line 212: shouldn't we use the logger here?

src/server/utils/logging/fastifyPlugin.js
```diff
@@ -200,12 +201,17 @@ const fastifyPlugin = (fastify, _opts, done) => {
   });
 
   fastify.addHook('onResponse', async (request, reply) => {
-    endTimer(request, $ResponseBuilder);
-    // same as 'reply.getResponseTime()' but our approach
-    // helps us to make the code cleaner
-    endTimer(request, $RequestFullDuration);
-
-    logClientRequest(request, reply);
+    try {
```
giulianok, line 204: why is the try/catch needed here? it catches the error, logs it, and throws it, meaning that we only need to log it right? Fastify handles uncaught errors for us (https://www.fastify.io/docs/latest/Reference/Errors/) so no need for a try/catch

src/server/utils/logging/fastifyPlugin.js
```diff
@@ -137,7 +138,7 @@ const logClientRequest = (request, reply) => {
   logger.info(configuredLog);
 };
 
-export const setConfigureRequestLog = (newConfigureRequestLog) => {
+export const setConfigureRequestLog = (newConfigureRequestLog = passThrough) => {
```
giulianok, line 141: why would somebody call `setConfigureRequestLog` without passing an argument?

giulianok (changes requested):

src/server/utils/logging/fastifyPlugin.js
JAdshead: its called everytime during onModule load

src/server/utils/logging/fastifyPlugin.js
JAdshead: this does use it. console is monkey patched to use the logger https://github.com/americanexpress/one-app/blob/main/src/server/utils/logging/setup.js#L28

src/server/utils/logging/fastifyPlugin.js
giulianok: because it shouldn't be called without an argument

JAdshead: It shouldn't but it is. I don't see the issue having a default to improve resiliency

giulianok: I won't block it but it's a bad practice. The function by starting with the word `set` lets the developer know that it "sets" something, and the data will always be expected (you cannot set something with no data, otherwise would be like a reset). That's the reason why all functions, like the `write` from `fs` will throw an error if data is not provided. If we want a set to be "safe" I would rename it to something like `safeSetConfigureRequestLog`, but, still, ideally, the user should always pass data

giulianok (approved):

PR #1066, opened 2023-08-02
fix(tracer): ending timers without starting would crash server

Matthew-Mallimo:
## Description

If you ended a timer without first starting it, it would throw an error.

## Motivation and Context

While testing the redirectAllowList, I found that if you throw an error within [createRequestHtmlFragment](https://github.com/americanexpress/one-app/blob/5.x.x/src/server/middleware/createRequestHtmlFragment.jsx) before starting serverPhaseTimer 7, the error would get caught and attempt to end timer 7. This would end up throwing an error outside of a try/catch and crash the server

## How Has This Been Tested?

Tested with the same method used to find this error, throwing an error in the redirect flow an ensuring the server didn't crash.

Also unit tested

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/src/server/utils/tracer.js
+++ b/src/server/utils/tracer.js
@@ -75,6 +75,10 @@ export class Tracer {
 
   // End a timer for a server phase
   endServerPhaseTimer = (serverPhasekey) => {
+    if (!this.#serverPhaseTimers[serverPhasekey]) {
+      console.warn('Tracer Error: endServerPhaseTimer was called without startServerPhaseTimer');
+      return;
+    }
     this.#serverPhaseTimers[serverPhasekey].endTimeNs = process.hrtime.bigint();
   }
 
@@ -87,12 +91,20 @@ export class Tracer {
 
   // End a timer for a fetch.
   endFetchTimer = (key) => {
+    if (!this.#fetchTimers[key]) {
+      console.warn('Tracer Error: endFetchTimer was called without startFetchTimer');
+      return;
+    }
     this.#fetchTimers[key].endTimeNs = process.hrtime.bigint();
   }
 
   // End a timer for a fetch with an error
   endFetchTimerWithError = (key, error) => {
     this.endFetchTimer(key);
+    if (!this.#fetchTimers[key]) {
+      console.warn('Tracer Error: endFetchTimerWithError was called without startFetchTimer');
+      return;
+    }
     this.#fetchTimers[key].error = error;
   }
 
--- a/__tests__/server/utils/tracer.spec.js
+++ b/__tests__/server/utils/tracer.spec.js
@@ -24,6 +24,7 @@ import {
 } from '../../../src/server/utils/tracer';
 
 jest.spyOn(console, 'log');
+jest.spyOn(console, 'warn');
 jest.useFakeTimers().setSystemTime(new Date('1993-07-25'));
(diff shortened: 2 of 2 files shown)


JAdshead (approved):

src/server/utils/tracer.js
```diff
@@ -75,7 +75,11 @@ export class Tracer {
 
   // End a timer for a server phase
   endServerPhaseTimer = (serverPhasekey) => {
-    this.#serverPhaseTimers[serverPhasekey].endTimeNs = process.hrtime.bigint();
+    if (this.#serverPhaseTimers[serverPhasekey]) {
+      this.#serverPhaseTimers[serverPhasekey].endTimeNs = process.hrtime.bigint();
+    } else {
+      console.warn('Tracer Error: endServerPhaseTimer was called without startServerPhaseTimer');
+    }
```
PixnBits, line 82: exit early?
```suggestion
    if (!this.#serverPhaseTimers[serverPhasekey]) {
      console.warn('Tracer Error: endServerPhaseTimer was called without startServerPhaseTimer');
      return;
    }
    this.#serverPhaseTimers[serverPhasekey].endTimeNs = process.hrtime.bigint();
```

src/server/utils/tracer.js
```diff
@@ -87,13 +91,21 @@ export class Tracer {
 
   // End a timer for a fetch.
   endFetchTimer = (key) => {
-    this.#fetchTimers[key].endTimeNs = process.hrtime.bigint();
+    if (this.#fetchTimers[key]) {
+      this.#fetchTimers[key].endTimeNs = process.hrtime.bigint();
+    } else {
+      console.warn('Tracer Error: endFetchTimer was called without startFetchTimer');
+    }
```
PixnBits, line 98:
```suggestion
    if (!this.#fetchTimers[key]) {
      console.warn('Tracer Error: endFetchTimer was called without startFetchTimer');
      return;
    }
    this.#fetchTimers[key].endTimeNs = process.hrtime.bigint();
```

src/server/utils/tracer.js
```diff
@@ -87,13 +91,21 @@ export class Tracer {
...
 
   // End a timer for a fetch with an error
   endFetchTimerWithError = (key, error) => {
     this.endFetchTimer(key);
-    this.#fetchTimers[key].error = error;
+    if (this.#fetchTimers[key]) {
+      this.#fetchTimers[key].error = error;
+    } else {
+      console.warn('Tracer Error: endFetchTimerWithError was called without startFetchTimer');
+    }
```
PixnBits, line 108:
```suggestion
    if (!this.#fetchTimers[key]) {
      console.warn('Tracer Error: endFetchTimerWithError was called without startFetchTimer');
      return;
    }
    this.#fetchTimers[key].error = error;
```

PixnBits (changes requested):

src/server/utils/tracer.js
PixnBits: it's a pattern from https://effectivejs.com/ but the book is at home so I don't have a link handy to the specific chapter, sorry 😓

PixnBits (approved):

JAdshead (approved):

PR #1067, opened 2023-08-04
feat(node): upgrade to node 18

Matthew-Mallimo:
## Description

Upgrade to Node 18

Had to add the flag `--dns-result-order=ipv4first` to node because of an issue with Node 17 where `localhost` gets resolved using IPv6.

https://github.com/nodejs/node/issues/40702#issue-1042587993

Also added the `--no-experimental-fetch` because we dont want to be using an experimental feature in production. Also we already polyfill fetch.

## Motivation and Context

Node 16 is EOL in September

## How Has This Been Tested?

Ran server locally, unit tests, integration tests.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/__tests__/server/utils/createCircuitBreaker.spec.js
+++ b/__tests__/server/utils/createCircuitBreaker.spec.js
@@ -132,13 +132,7 @@ describe('Circuit breaker', () => {
     setEventLoopDelayThreshold(-1);
     jest.advanceTimersByTime(5e3 + 10);
     await mockCircuitBreaker.fire('hola, mundo');
-    expect(consoleErrorSpy.mock.calls).toMatchInlineSnapshot(`
-      [
-        [
-          [Error: Opening circuit, p(100) event loop delay (0ms) is > eventLoopDelayThreshold (-1ms)],
-        ],
-      ]
-    `);
+    expect(consoleErrorSpy).toHaveBeenCalled();
   });
 
   it('should log when the circuit opens', () => {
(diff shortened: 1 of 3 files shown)


__tests__/server/utils/createCircuitBreaker.spec.js
```diff
@@ -132,13 +132,7 @@ describe('Circuit breaker', () => {
...
     jest.advanceTimersByTime(5e3 + 10);
     await mockCircuitBreaker.fire('hola, mundo');
-    expect(consoleErrorSpy.mock.calls).toMatchInlineSnapshot(`
-      [
-        [
-          [Error: Opening circuit, p(100) event loop delay (0ms) is > eventLoopDelayThreshold (-1ms)],
-        ],
-      ]
-    `);
+    expect(consoleErrorSpy).toHaveBeenCalled();
```
Matthew-Mallimo, line 135: Had to change this because the snapshot kept coming back as :

> [Error: Opening circuit, p(100) event loop delay (0.005112ms) is > eventLoopDelayThreshold (-1ms)],

Seems like a rounding difference, but didn't want to spend too much time figuring that out

10xLaCroixDrinker (dismissed):

__tests__/server/utils/createCircuitBreaker.spec.js
Matthew-Mallimo: Resolved in b6a21963f40257c3459ce0fe282994c5a62dbbd0 by mocking perf_hooks

JAdshead (approved):

bishnubista (approved):

PR #1085, opened 2023-08-14
fix(image): exclude devDeps from non-dev image

PixnBits:
## Description

move devDeps if-env, node-fetch, and proxy-agent from deps list

## Motivation and Context

proxy-agent was found in the non-dev image when it shouldn't be there, so a search through the dependencies was started
all `NODE_ENV==='development'` was set to `false` and subsequent code simplification performed
using the dependencies list, using `$ grep -rl $dependency src` showed that there were a few not used:
* abort-controller
* create-shared-react-context
* if-env
* node-fetch
* prop-types
* proxy-agent
* reselect
* url-polyfill

Of these, some were found to be used by the bundler as dependencies that the one-app runtime provides to modules ([webpack/webpack.common.js](https://github.com/americanexpress/one-app-cli/blob/v6.12.0/packages/one-app-bundler/webpack/webpack.common.js#L102-L155), [webpack/app/webpack.client.js](https://github.com/americanexpress/one-app-cli/blob/ab9e6de681faf126642814c168a780109a0c3483/packages/one-app-bundler/webpack/app/webpack.client.js#L69)). This left if-env, node-fetch, and proxy-agent; the latter two are used by the development CDN. if-env is used in a package.json script which is similarly only used during development (note that the Dockerfile does **not** use package.json scripts, e.g. `$ npm start` as the CMD or ENTRYPOINT, so as to avoid unnecessary additional process(es))

## How Has This Been Tested?

commit hook, awaiting PR checks

## Types of Changes

- [x] Dependency update
- [x] Security update

## Checklist:

smaller images are faster to transfer and easier to store (use less bits on disk)


Matthew-Mallimo (approved):

bishnubista (approved):

package.json
```diff
@@ -184,11 +182,13 @@
     "lodash.set": "^4.3.2",
     "markdown-table": "^3.0.2",
     "mkdirp": "^2.1.5",
+    "node-fetch": "^2.6.7",
```
JAdshead, line 185: maybe node-fetch can be removed entirely as cross-fetch is also used.

JAdshead (approved):

package.json
PixnBits: I had that thought too, but I think that might be a bigger scope (so different PR) due to usage being in the dev CDN which is in Node.js so being specific might be useful? Not 100% sure either way

PR #1090, opened 2023-08-14
feat(Dockerfile): use .nvmrc to build the image

PixnBits:
## Description

use .nvmrc to build the image
and run workflows, creating a single point to define the Node.js version to use and thus making upgrades easier

## Motivation and Context

Avoid the myriad of locations where Node.js versions are recorded, i.e. Dockerfile and workflows. package.json engines will still need manual updates though (perhaps that can be a workflow, or a unit test to ensure it is satisfied by the contents of `.nvmrc`?)

## How Has This Been Tested?

built locally, though not run, but did see the expected label used when manually running `$ docker build . --build-arg VERSION=18.17.0` and `docker build . --build-arg VERSION=$(cat .nvmrc)`

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

faster Node.js version updates? this is more for the maintainers so no impact to devs is the basic requirement 😅


__tests__/package.spec.js
```diff
@@ -13,16 +13,69 @@
...
+        expect.assertions(1);
+        const rawNVM = await fs.readFile(path.join(PACKAGE_DIR_PATH, '.nvmrc'), 'utf8');
+        const { engines } = JSON.parse(await fs.readFile(path.join(PACKAGE_DIR_PATH, 'package.json')));
+        expect(satisfies(rawNVM, engines.node)).toBe(true);
+      });
+      it('is inline with .nvmrc', async () => {
+        expect.assertions(1);
+        const rawNVM = await fs.readFile(path.join(PACKAGE_DIR_PATH, '.nvmrc'), 'utf8');
+        const { engines } = JSON.parse(await fs.readFile(path.join(PACKAGE_DIR_PATH, 'package.json')));
+        expect(engines.node).toBe(`^${parse(rawNVM).major}`);
```
PixnBits, line 77: proof of work: https://github.com/americanexpress/one-app/actions/runs/5869749612/job/15915395595

do we want to keep the original `>=` or is `^` better? (my thought is with this check we'll be updating the field alongside the `.nvmrc` file, so shouldn't be an impediment to intentionally using a different major version of Node.js)

code-forger (approved):

code-forger (approved):

code-forger (changes requested):
See comment

code-forger:
This PR will look like its not building, because our settings require workflows that this PR removes.

Once this PR is ready and approved, the setting should be changed immediately before merge

.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml
```diff
@@ -9,9 +9,6 @@ jobs:
   one-app-unit-tests:
     name: One App Unit and Lint Tests
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        node-version: [16.x, "^18 < 18.16"]
```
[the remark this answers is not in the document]
dogpatch626: +1

.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml
```diff
@@ -22,7 +19,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v3
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version-file: .nvmrc
```
JAdshead, line 22: lets revert this change and above, apart from that LGTM!!
```suggestion
          node-version: ${{ matrix.node-version }}
```

.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml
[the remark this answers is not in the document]
PixnBits: I was hoping to find a way to build that matrix with the .nvmrc file version, but that can be a different PR (if I do figure that out 😅)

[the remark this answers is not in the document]
PixnBits: changed in 527638f

.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml
PixnBits: reverted in 527638f 🙇

JAdshead (dismissed):

.nvmrc
```diff
@@ -0,0 +1 @@
+18.17.0
```
PixnBits, line 1: after #1087 this should be
```suggestion
18.17.1
```

PixnBits: updated in 9e5b850

JAdshead (dismissed):

.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [16.x, "^18 < 18.16"]
+        node-version: [16.x, 18.x]
```
PixnBits, line 14: from https://github.com/americanexpress/one-app/actions/runs/6042820335/job/16398714236?pr=1090
perhaps this should instead be
```suggestion
        node-version: [18.x]
```
until 20.x is LTS?

JAdshead: maybe revert the engines field. This still supports 16 for another week or so

PixnBits: whoops, forgot that change was me, good catch!

PixnBits: reverted the engines and tests of it in c63a856

JAdshead (dismissed):

dogpatch626 (dismissed):

Dockerfile
```diff
@@ -1,8 +1,9 @@
+ARG VERSION=latest
```
Matthew-Mallimo, line 1: will this not bring in node 20?

PixnBits: It will bring in non-LTS versions too, if the build argument is not provided (see `docs/guides/Running-In-Production.md`). Many of the changes in this PR are to supply that argument 👍

PixnBits: We could use a specific version as the default, instead of `latest` but that could be indirection and cause surprises? 🤔

PixnBits: From https://hub.docker.com/_/node there is an `lts` tag, we could use that as the default?
```suggestion
ARG VERSION=lts
```

JAdshead (approved):

dogpatch626 (approved):

PR #1099, opened 2023-08-29
feat(styleLoader): aggregate stylesheets and dedupe if already loaded

eddhurst:
## Description

This PR:

- fixes a cascade ordering issue between server and client bundles, leading to style inconsistencies when hydrated.
- enables deduplication between modules for identical styles, i.e. if a component library imports styles via css modules.
- provides the ability for SSR generated styles to directly match the expected output of CSR styles, preventing a flash of restyling as the browser processes the Recalculate Styles task.

## Motivation and Context

When using a component library (or any dependency) that imports styles via CSS Modules, the bundler output generates a server bundle that is inconsistent with the client side bundle, as well as resulting in it being difficult to provide style overrides without resorting to CSS hacks (like !important).

In order to resolve this and match the client side bundle output it's necessary to split the styles into multiple style tags, and then order them to allow for cascade overrides. However, the OneApp styles loader currently wraps all SSR styles in a single style tag.

This change retains the existing functionality in order to provide backwards compatibility for modules generated with older bundlers, whilst allowing for bundlers to provide a new key: `aggregatedStyles`.

## How Has This Been Tested?

Generated a root module and three child modules that share a component library dependency, with each child module rendering two instances of the same component. The first one unchanged, the second with style overrides applied.

Multiple modules helped to prove that deduplication happened as intended, with only a single instance of the component library styles being rendered, with overrides applying with the expected cascade priority.

This was additionally run with an older version of OneApp, and an older bundle to prove backwards compatibility is unaffected.

Screenshot 2023-08-29 at 16 10 37

Screenshot 2023-08-29 at 22 30 17

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

This allows developers to provide overrides to styles imported via CSS modules within their applications without resorting to CSS hacks and forcing specificity.

### DIFF
--- a/__tests__/server/utils/__snapshots__/renderModuleStyles.spec.js.snap
+++ b/__tests__/server/utils/__snapshots__/renderModuleStyles.spec.js.snap
@@ -1,7 +1,49 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renderModuleStyles should handle a mix of modules with and without SSR styles 1`] = `"<style class="ssr-css">.test-module-b_class { color: red; }</style><style class="ssr-css">.test-module-d_class { color: red; }</style>"`;
+exports[`renderModuleStyles should deduplicate aggregatedStyles that have the same digest hash 1`] = `
+"
+<style id="shared_hash_deps" data-ssr="true">
+  .test-module-a_deps { color: blue; }
+</style>
+<style id="shared_hash_local" data-ssr="true">
+  .test-module-a_local { color: rebeccapurple; }
+</style>
+<style id="unique-hash_deps" data-ssr="true">
+  .test-module-c_deps { color: blue; }
+</style>
+<style id="unique-hash_local" data-ssr="true">
+  .test-module-c_local { color: rebeccapurple; }
+</style>"
+`;
+
+exports[`renderModuleStyles should handle a mix of modules with and without SSR styles 1`] = `
+"<style class="ssr-css">.test-module-b_class { color: red; }</style><style class="ssr-css">.test-module-d_class { color: red; }</style>
+<style id="test-module-e_deps" data-ssr="true">
+  .test-module-e_deps { color: blue; }
+</style>
+<style id="test-module-e_local" data-ssr="true">
+  .test-module-e_local { color: rebeccapurple; }
+</style>"
+`;
 
 exports[`renderModuleStyles should handle modules with SSR styles 1`] = `"<style class="ssr-css">.test-module_class { color: red; }</style>"`;
 
-exports[`renderModuleStyles should not send empty styles 1`] = `"<style class="ssr-css">.test-module-b_class { color: red; }</style>"`;
+exports[`renderModuleStyles should handle modules with aggregated SSR styles 1`] = `
+"
+<style id="test-module_deps" data-ssr="true">
+  .test-module_deps { color: blue; }
+</style>
+<style id="test-module_local" data-ssr="true">
+  .test-module_local { color: rebeccapurple; }
+</style>"
+`;
+
+exports[`renderModuleStyles should not send empty styles 1`] = `
+"<style class="ssr-css">.test-module-b_class { color: red; }</style>
+<style id="test-module-d_deps" data-ssr="true">
+  .test-module-d_deps { color: blue; }
+</style>
+<style id="test-module-d_local" data-ssr="true">
+  .test-module-d_local { color: rebeccapurple; }
+</style>"
+`;
--- a/__tests__/server/utils/__snapshots__/renderModuleStyles.spec.js.snap
+++ b/__tests__/server/utils/__snapshots__/renderModuleStyles.spec.js.snap
@@ -1,49 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renderModuleStyles should deduplicate aggregatedStyles that have the same digest hash 1`] = `
-"
-<style id="shared_hash_deps" data-ssr="true">
-  .test-module-a_deps { color: blue; }
-</style>
-<style id="shared_hash_local" data-ssr="true">
-  .test-module-a_local { color: rebeccapurple; }
-</style>
-<style id="unique-hash_deps" data-ssr="true">
-  .test-module-c_deps { color: blue; }
-</style>
-<style id="unique-hash_local" data-ssr="true">
-  .test-module-c_local { color: rebeccapurple; }
-</style>"
-`;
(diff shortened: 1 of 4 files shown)


src/server/utils/renderModuleStyles.js
```diff
@@ -16,12 +16,58 @@
...
+</style>`;
+
+const generateServerStyleTag = (css) => `<style class="ssr-css">${css}</style>`;
+
+const filterOutDuplicateDigests = (sheet, existingDigests) => sheet
+  .filter(({ css, digest }) => css && digest && !existingDigests.has(digest))
+  .map((styles) => {
+    existingDigests.add(styles.digest);
+    return styles;
+  });
```
JAdshead, line 35:
```suggestion
```

src/server/utils/renderModuleStyles.js
```diff
@@ -16,12 +16,58 @@
...
+
+      const { aggregatedStyles } = module.ssrStyles;
+      return {
+        ...acc,
+        aggregated: [
+          ...acc.aggregated,
+          ...filterOutDuplicateDigests(aggregatedStyles, existingDigests),
+        ],
+      };
+    }, { aggregated: [], legacy: [] });
```
JAdshead, line 62: rough outline on how to remove side effect from function.

```suggestion
      const { aggregatedStyles } = module.ssrStyles;
      const uniqStyles = filterOutDuplicateDigests(aggregatedStyles, existingDigests)
      uniqStyles.forEach(style => existingDigests.add(style.digest));
      return {
        ...acc,
        aggregated: [
          ...acc.aggregated,
          ...uniqStyles
        ],
      };
    }, { aggregated: [], legacy: [] });
```

__tests__/server/utils/__snapshots__/renderModuleStyles.spec.js.snap
```diff
@@ -1,7 +1,49 @@
...
+<style id="unique-hash_local" data-ssr="true">
+  .test-module-c_local { color: rebeccapurple; }
+</style>"
+`;
+
+exports[`renderModuleStyles should handle a mix of modules with and without SSR styles 1`] = `
+"<style class="ssr-css">.test-module-b_class { color: red; }</style><style class="ssr-css">.test-module-d_class { color: red; }</style>
+<style id="test-module-e_deps" data-ssr="true">
+  .test-module-e_deps { color: blue; }
+</style>
```
JAdshead, line 23: could both be consistently formatted ?

src/server/utils/renderModuleStyles.js
```diff
@@ -16,12 +16,58 @@
...
 import { getModule } from 'holocron';
 
+/**
+ * Generates a style tag with unique ID attribute per CSS file loaded.
+ * @returns {string}
+ */
+const generateStyleTag = ({ css, digest }) => `
+<style id="${digest}" data-ssr="true">
+  ${css}
+</style>`;
```
JAdshead, line 26: keep the file size as small as possible
```suggestion
const generateStyleTag = ({ css, digest }) => `<style id="${digest}" data-ssr="true">${css}</style>`;
```

JAdshead (changes requested):
Cool feature. apart from the small refactoring looks good!!

src/server/utils/renderModuleStyles.js
```diff
@@ -16,12 +16,58 @@
...
+
+const generateServerStyleTag = (css) => `<style class="ssr-css">${css}</style>`;
+
+const filterOutDuplicateDigests = (sheet, existingDigests) => sheet
+  .filter(({ css, digest }) => css && digest && !existingDigests.has(digest))
+  .map((styles) => {
+    existingDigests.add(styles.digest);
```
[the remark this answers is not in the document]
eddhurst: [4eaf62c](https://github.com/americanexpress/one-app/pull/1099/commits/4eaf62c6c1b600055ca54c1436f50f7b7f6e372a)

src/server/utils/renderModuleStyles.js
eddhurst: [aeae31c](https://github.com/americanexpress/one-app/pull/1099/commits/aeae31ca6e079b536f3481399044c2072942391a)

__tests__/server/utils/__snapshots__/renderModuleStyles.spec.js.snap
eddhurst: [aeae31c](https://github.com/americanexpress/one-app/pull/1099/commits/aeae31ca6e079b536f3481399044c2072942391a)

src/server/utils/renderModuleStyles.js
```diff
@@ -16,12 +16,58 @@
...
+
+  const collatedStyles = modulesWithSSRStyles
+    .reduce((acc, module) => {
+      // Backwards compatibility for older bundles.
+      if (!module.ssrStyles.aggregatedStyles) {
+        const ssrStylesFullSheet = module.ssrStyles.getFullSheet();
+        if (ssrStylesFullSheet) {
+          acc.legacy.push({ css: ssrStylesFullSheet });
+        }
+        return acc;
```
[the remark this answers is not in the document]
eddhurst: [aeae31c](https://github.com/americanexpress/one-app/pull/1099/commits/aeae31ca6e079b536f3481399044c2072942391a)

eddhurst:
> Cool feature. apart from the small refactoring looks good!!

Thanks! and thanks for the feedback, addressed all comments

Matthew-Mallimo (approved):

__tests__/server/utils/__snapshots__/renderModuleStyles.spec.js.snap
```diff
@@ -1,7 +1,11 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renderModuleStyles should handle a mix of modules with and without SSR styles 1`] = `"<style class="ssr-css">.test-module-b_class { color: red; }</style><style class="ssr-css">.test-module-d_class { color: red; }</style>"`;
+exports[`renderModuleStyles should deduplicate aggregatedStyles that have the same digest hash 1`] = `"<style id="shared_hash_deps" data-ssr="true">.test-module-a_deps { color: blue; }</style><style id="shared_hash_local" data-ssr="true">.test-module-a_local { color: rebeccapurple; }</style><style id="unique-hash_deps" data-ssr="true">.test-module-c_deps { color: blue; }</style><style id="unique-hash_local" data-ssr="true">.test-module-c_local { color: rebeccapurple; }</style>"`;
```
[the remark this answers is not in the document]
eddhurst: strictly speaking none - although a style tag ought not have a class name because we're not applying styles onto the element. The idea here was more of a debug step to allow us to more easily track which styles are being hydrated and which are not. We could dispense with this entirely, but it felt like that might still be a useful step for more easily debugging / visualising the changes.

JAdshead (approved):

code-forger (approved):

PR #1101, opened 2023-08-30
feat(*): add one app debug port to runtime and expose

drewcur:
Acknowledge the debugging port initially added in one-app-runner and expose in the Dockerfile to access.

## Description
To fully permit the [use-debug](https://github.com/americanexpress/one-app-cli/tree/main/packages/one-app-runner#use-debug-optional) feature to work, first acknowledge the env var setting to set the foundation and second, expose the port in the Dockerfile so it can be used when containerized.

## Motivation and Context
Currently, the [use-debug](https://github.com/americanexpress/one-app-cli/tree/main/packages/one-app-runner#use-debug-optional) feature doesn't work because the port isn't expose

## How Has This Been Tested?
See comments.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

Should be able to connect the node debugger through the Docker container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,7 @@ EXPOSE 3000
 EXPOSE 3001
 EXPOSE 3002
 EXPOSE 3005
+EXPOSE 9229
 WORKDIR /opt/one-app
 RUN chown node:node /opt/one-app
 USER $USER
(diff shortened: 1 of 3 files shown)


drewcur:
Still working on trying to build a local docker image, would it be a requirement to have tested it locally? Is there any other way I can validate these changes?

JAdshead:
I believe that the changes to https://github.com/americanexpress/one-app-cli/pull/559 should be enough to have this enabled.

PR #1103, opened 2023-08-31
chore(docker): disable docker buildkit

Matthew-Mallimo:
Disable docker buildkit to be consistent with our previous images


PixnBits:
Did the integration tests run with the BuildKit or legacy builder built image? 🤔

PR #1166, opened 2023-10-31
fix(reactHtml): renderScripts adding unknown integrity

JAdshead:
## Description

Fallback externals do not support legacy, remove legacy logic to ensure does not load missing scripts.
Don't attempt to include integrity when missing, this is requires as fallback externals do not include integrity values

## Motivation and Context

Fix SSR when fallback externals are required and has no integrity value provided

## How Has This Been Tested?

Test suite

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "fastify-plugin": "^4.5.1",
     "flat": "^5.0.2",
     "helmet": "^7.0.0",
-    "holocron": "^1.9.0",
+    "holocron": "^1.9.1",
     "holocron-module-route": "^1.7.0",
     "immutable": "^4.1.0",
     "ip": "^1.1.8",
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
     "fastify-plugin": "^4.5.1",
     "flat": "^5.0.2",
     "helmet": "^7.0.0",
-    "holocron": "^1.8.2",
+    "holocron": "^1.9.0",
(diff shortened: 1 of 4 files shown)


Matthew-Mallimo:
> fallback externals do not include integrity values
But shouldn't they?

JAdshead:
> > fallback externals do not include integrity values
> 
> But shouldn't they?

Yes, but instead of crashing in the client this will more gracefully load the script

package.json
```diff
@@ -112,7 +112,7 @@
     "fastify-plugin": "^4.5.1",
     "flat": "^5.0.2",
     "helmet": "^7.0.0",
-    "holocron": "^1.8.2",
```
JAdshead, line 115: This will need to be updated to next release with holocron updates

dogpatch626 (approved):

package.json
```diff
@@ -112,7 +112,7 @@
...
+    "holocron": "^1.9.1",
```
Matthew-Mallimo, line 115: I this required? If I remember correctly the 1.9.0 release was done unintentionally?

JAdshead: yes, it includes the fixes for integrity loading in holocron

Matthew-Mallimo (approved):

dogpatch626 (approved):

PR #1194, opened 2023-11-27
fix(runTime): use relative urls for defaulted report variables

### LINKED ISSUE #1193: Default ONE_CLIENT_REPORTING_URL value does not work in Docker
# 🐞 Bug Report

## Describe the bug

The default value for the ONE_CLIENT_REPORTING_URL is using the IP address that node sees.  When running inside docker, this is an internal IP which is not exposed.  For example: `http://172.17.0.2:3000/_/report/errors`. As a result, the error reporting fails.

This also applies to ONE_CLIENT_CSP_REPORTING_URL.

## To Reproduce

1. Start One App inside docker using one-app-runner.
2. Trigger a javascript error, for instance by using an undefined variable.
3. See call to http://172.17.0.2:3000/_/report/errors hang and then timout.

## Expected behavior

Error reporting call should succeed and Javascript error should be reported in console logs.

## System information

- OS: macOS
- Version of one-app: 6.8.2
- Are you using building one-app from source or using the release Docker image? Docker

smackfu:
## Description
Changes the default development ONE_CLIENT_REPORTING_URL and ONE_CLIENT_CSP_REPORTING_URL to use relative URLs instead of URLs with a hostname.

## Motivation and Context
Fixes #1193 by not using the internal docker IP address.

## How Has This Been Tested?
Ran locally with one-app-runner using these updated variables and verified that report endpoints responded with 204 and printed reports to console.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

When running in Docker, Javascript errors should be printed to the console rather than not being reported.

### DIFF
--- a/__tests__/server/config/env/runTime.spec.js
+++ b/__tests__/server/config/env/runTime.spec.js
@@ -357,8 +357,7 @@ describe('runTime', () => {
 
     it('has a default value for development', () => {
       process.env.NODE_ENV = 'development';
-      expect(clientReportingUrl.defaultValue()).toBeDefined();
-      expect(clientReportingUrl.defaultValue()).toMatch(/^https?:\/\//);
+      expect(clientReportingUrl.defaultValue()).toBe('/_/report/errors');
     });
 
     it('has no default value for production', () => {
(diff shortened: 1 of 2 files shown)


JAdshead (approved):

smackfu:
Interestingly the documentation already says this is what is being used for the values:
https://github.com/americanexpress/one-app/blob/main/docs/api/server/Environment-Variables.md

Screenshot 2023-11-27 at 1 30 16 PM

10xLaCroixDrinker (approved):

PR #1210, opened 2023-12-07
fix(devCDN): handle different module files

code-forger:
## Motivation and Context
With the addition of module-config.json, holocron now loads two files per module. These files were triggering the simple 'remove duplicates' functionality. This was the cause of the 'removed file from cache' spam on startup, and effectively neutralised the cache as every file was always getting wiped.

Instead, now consider every part of the url except the version when detecting conflicts, allowing for two different file names to be treated differently

## How Has This Been Tested?
run against multiple tenancies
unit tests

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

Faster app startup, less log spam


src/server/utils/cdnCache.js
```diff
@@ -89,15 +90,21 @@ export const writeToCache = (content, delay = 500) => {
...
 
-  const updatedCachedModules = cachedModules;
-  Object.keys(updatedCachedModules).forEach((cachedModuleKey) => {
-    if (cachedModuleKey.match(new RegExp(`\\b\\/${matchingModule}\\/\\b`))) {
-      delete updatedCachedModules[cachedModuleKey];
-      console.log(`Deleted ${cachedModuleKey} from cache`);
-    }
-  });
+export const removeExistingEntryIfConflicting = (url, cachedModuleFiles) => {
+  const updatedCachedModules = cachedModuleFiles;
```
JAdshead, line 100: delete will remove matching module from both `updatedCachedModules` and `cachedModuleFiles`. so the test in https://github.com/americanexpress/one-app/pull/1210/files#diff-0264af67b0d133bfc5c2b5d595e25c416dbc5e86ed199e8cd64d9924d99b2126L245 will always be true even if it removes files it should not. 
 
```suggestion
  const updatedCachedModules = { ...cachedModuleFiles };
```

code-forger: [8a673e6](https://github.com/americanexpress/one-app/pull/1210/commits/8a673e65b815cf25e7000186e92a184ee7ac22a1)

Addressed

10xLaCroixDrinker (approved):

bishnubista (approved):

PR #1212, opened 2023-12-13
feat(react): update to react 18

JAdshead:
BREAKING CHANGE: major version to React

## Description

Update React from 17 to 18.

## Types of Changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Dependency update
- [ ] Security update

## Checklist:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG VERSION=lts
+ARG VERSION=18
```
giulianok, line 1: Is this hardcoded because LTS is now 20?

__tests__/client/initClient.spec.js
```diff
@@ -153,8 +153,8 @@ describe('initClient', () => {
 
     await initClient();
```
giulianok, line 155: Shouldn’t we update `initClient` as well to start using the new react approach?

giulianok (changes requested):

__tests__/client/initClient.spec.js
JAdshead: This is using the new approach. https://github.com/americanexpress/one-app/pull/1212/files#diff-cb26e299c3c333198ccd95e53e4f18369a1268c72074277b91d0819b309f122bR73-R79

If you are referring to renderToPipeableStream thats a huge rewrite that would be part of another PR. its also not compatible with Fastify out of the box.

Dockerfile
JAdshead: yes, this is just for the fallback. All uses currently use the version specified in `.nvmrc`.

giulianok: if it's a fallback why not leave it as `lts`?

__tests__/client/initClient.spec.js
giulianok: gotcha

Dockerfile
JAdshead: because node 20 fails.

dogpatch626 (approved):

prod-sample/sample-modules/frank-lloyd-root/0.0.0/src/components/FrankLloydRoot.jsx
```diff
@@ -34,7 +34,7 @@ export function FrankLloydRoot({ children, config }) {
           { rel: 'icon', href: 'https://sample-cdn.frank/favicon.ico' },
         ]}
       />
-      <pre className="value-provided-from-config">{ config.get('someApiUrl') }</pre>
+      <pre suppressHydrationWarning={true} className="value-provided-from-config">{ config.get('someApiUrl') }</pre>
```
Matthew-Mallimo, line 37: are we not able to hydrate this properly for some reason?

JAdshead: the tests validate different values on client and ssr so we deliberately cause a mismatch, react 17 was happy with the react 18 not so much

giulianok: curious about those values, why do we validate different values on client and server?

test-setup.js
```diff
@@ -17,8 +17,13 @@
...
 import 'core-js/stable';
 import 'regenerator-runtime/runtime';
+import { TextEncoder, TextDecoder } from 'util';
 
 process.env.ONE_CONFIG_ENV = 'test';
 
 // Required for logger tests
 process.setMaxListeners(15);
+
+// Issue with Jest this could be moved to amex-jest-preset-react
```
giulianok, line 27: what's the issue about?

Dockerfile
giulianok: can you provide more info? i have been testing node 20 and haven't find any issues so far, could help me find any problems

prod-sample/sample-modules/frank-lloyd-root/0.0.0/src/components/FrankLloydRoot.jsx
JAdshead: It validates that the client state config is different from the server state config.

Dockerfile
JAdshead: you will need to run the integration tests with node 20. This also locks the version used to a node version which is supported. LTS changes ideally this should not end up in a broken state each time a new version becomes LTS

test-setup.js
JAdshead: `jest-environment-jsdom` does not provide TextEncode and TextDecoder globals. This result in a `"ReferenceError: TextEncoder is not defined"` when running tests.

JAdshead, line 27:
```suggestion
// Jest does not provide TextEncoder & TextDecoder on global. This could be moved to amex-jest-preset-react
```

Dockerfile
giulianok: thanks for the context. Yeah that's true, better to be safe by locking the version and upgrade as soon as we test newer versions

giulianok (approved):

dogpatch626 (approved):

Matthew-Mallimo (approved):

PR #1213, opened 2023-12-14
refactor(lumberjack): replace with util fn

JAdshead:
## Description

This is almost a direct lift and shift from https://github.com/americanexpress/lumberjack/blob/main/src/monkeypatches/attachHttpsRequestSpy.js.
For simplicity i combined the attach http and https request spies into single function.

## Motivation and Context

One App is now only using small part of Lumberjack. This will allow lumberjack to be deprecated. 

## How Has This Been Tested?

Test suite.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


src/server/utils/logging/setup.js
```diff
@@ -69,12 +69,4 @@ function outgoingRequestEndSpy(externalRequest, parsedUrl) {
...
 
-// In Node.js v8 and earlier https.request internally called http.request, but this is changed in
-// later versions
-// https://github.com/nodejs/node/blob/v6.x/lib/https.js#L206
-// https://github.com/nodejs/node/blob/v8.x/lib/https.js#L239
-// https://github.com/nodejs/node/blob/v10.x/lib/https.js#L271
-if (Number.parseInt(/^v(\d+)/.exec(process.version)[1], 10) > 8) {
-  monkeypatches.attachHttpsRequestSpy(outgoingRequestSpy, outgoingRequestEndSpy);
-}
-monkeypatches.attachHttpRequestSpy(outgoingRequestSpy, outgoingRequestEndSpy);
```
JAdshead, line 80: this has been simplified as One App has long since dropped support for node 8

src/server/utils/logging/attachSpy.js
```diff
@@ -0,0 +1,47 @@
+/*
```
JAdshead, line 1: https://github.com/americanexpress/lumberjack/blob/main/src/monkeypatches/attachSpy.js

src/server/utils/logging/attachRequestSpies.js
```diff
@@ -0,0 +1,83 @@
...
+
+    requestSpy(clientRequest, parsedUrl);
+
+    if (socketCloseSpy) {
+      onFinished(clientRequest, () => socketCloseSpy(clientRequest, parsedUrl));
+    }
+  };
+}
+
+export default function attachRequestSpies(requestSpy, socketCloseSpy) {
```
JAdshead, line 71: https://github.com/americanexpress/lumberjack/blob/main/src/monkeypatches/attachHttpRequestSpy.js

src/server/utils/logging/attachRequestSpies.js
```diff
@@ -0,0 +1,83 @@
...
+    throw new TypeError(`requestSpy must be a function (was "${typeof requestSpy}")`);
+  }
+
+  if (socketCloseSpy && (typeof socketCloseSpy !== 'function')) {
+    throw new TypeError(
+      `socketCloseSpy must be function if provided (was "${typeof socketCloseSpy}")`);
+  }
+
+  attachSpy(https, 'request', httpSpy('https', requestSpy, socketCloseSpy));
+  attachSpy(http, 'request', httpSpy('http', requestSpy, socketCloseSpy));
```
JAdshead, line 82: this is the only change. instead of having a function for each attachment we have a single function for attaching both.

src/server/utils/logging/attachRequestSpies.js
```diff
@@ -0,0 +1,83 @@
...
+  // some data is not stored in the clientRequest, have to duplicate some parsing
+  // adapted from https://github.com/nodejs/node/blob/894203dae39c7f1f36fc6ba126bb5d782d79b744/lib/_http_client.js#L22
+  if (typeof options === 'string') {
+    return url.parse(options);
+  }
+
+  return url.parse(url.format(buildUrlObject(options, defaultProtocol)));
+}
+
+function httpSpy(defaultProtocol, requestSpy, socketCloseSpy) {
```
JAdshead, line 57: https://github.com/americanexpress/lumberjack/blob/main/src/monkeypatches/httpSpy.js

src/server/utils/logging/attachRequestSpies.js
```diff
@@ -0,0 +1,83 @@
...
+    path: parsedPath.path,
+    pathname: parsedPath.pathname,
+    query: parsedPath.query,
+    search: parsedPath.search,
+  };
+  if (
+    (protocol === 'http:' && urlObject.port === 80)
+    || (protocol === 'https:' && urlObject.port === 443)
+  ) {
+    delete urlObject.port;
```
giulianok, line 42: `delete` keyword has a bad performance, can we unset the `port` by assigning `null` or `undefined`???

package.json
Matthew-Mallimo, line 1: Is there still more to bring over from lumberjack?

src/server/utils/logging/attachRequestSpies.js
```diff
@@ -0,0 +1,89 @@
...
+
+import https from 'node:https';
+import http from 'node:http';
+import url from 'node:url';
+import onFinished from 'on-finished';
+
+import attachSpy from './attachSpy';
+
+function buildUrlObject(options, defaultProtocol) {
+  // TODO: url.parse is deprecated, use new URL() instead
```
JAdshead, line 25: this change is beyond the scope of this PR

src/server/utils/logging/attachRequestSpies.js
```diff
@@ -0,0 +1,89 @@
...
+    search: parsedPath.search,
+  };
+  if (
+    (protocol === 'http:' && urlObject.port === 80)
+    || (protocol === 'https:' && urlObject.port === 443)
+  ) {
+    urlObject.port = undefined;
+  }
+  return urlObject;
+}
```
giulianok, line 51: (ref https://github.com/americanexpress/one-app/pull/1213#discussion_r1428884493), knowing that we can use `undefined` for `port`, then there's no need for the if statement + mutation, we could simplify it like: 

```suggestion
  const urlObject = {
    auth: options.auth,
    hostname: options.hostname || options.host || 'localhost',
    protocol,
    portx: options.port || (protocol === 'http:' ? 80 : 443),
    port: options.port && (
        (protocol === 'http:' && options.port !== 80) ||
        (protocol === 'https:' && options.port !== 443)
      ) ? options.port : undefined,
    hash: parsedPath.hash,
    path: parsedPath.path,
    pathname: parsedPath.pathname,
    query: parsedPath.query,
    search: parsedPath.search,
  };

  return urlObject;
}
```

The idea is that, if the `options.port` is provided then we validate and make sure it's a different port (different than `80` for `http` and different than `443` for `https`), otherwise we just set it to `undefined`.

Not, this leads me to the following question, why do we need to unset the port for those cases? can't just use `80` for `http` since that's the provided port? looks like that we need it's just `port: options.port`

JAdshead: This is not quiet the same result and I think having the removal as a separate conditional is clearer. I suspect this logic is based around old Node edge cases. 
This PR is intended as a simple lift and shift. This will need a rewrite as part of the removal `url.parse`

10xLaCroixDrinker:
commit type should be `refactor`

JAdshead:
> commit type should be `refactor`

will change on merge

__tests__/server/utils/logging/attachRequestSpies.spec.js
```diff
@@ -0,0 +1,148 @@
...
+      const requestSpy = jest.fn();
+      attachRequestSpies(requestSpy);
+      expect(attachSpy).toHaveBeenCalledTimes(2);
+
+      const httpsSpy = attachSpy.mock.calls[0][2];
+      const httpSpy = attachSpy.mock.calls[1][2];
+
+      const callOriginal = jest.fn(() => 'client request object');
+
+      httpsSpy(['http://example.tld'], callOriginal);
```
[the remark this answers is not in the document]
JAdshead: http is mocked - https://github.com/americanexpress/one-app/pull/1213/files#diff-db279ba2c5eab61489a06a64737477c9c45f7677994dcd2850782d32d30cc712R22

Matthew-Mallimo (approved):

__tests__/server/utils/logging/attachRequestSpies.spec.js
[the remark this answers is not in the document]
JAdshead: we could, these tests rely on attachSpy being mocked so they will all need a rewrite.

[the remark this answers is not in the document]
JAdshead: [f3c5079](https://github.com/americanexpress/one-app/pull/1213/commits/f3c5079450094df873a944bd4291356a8c251ee7)

Matthew-Mallimo (approved):

10xLaCroixDrinker (approved):

giulianok (approved):

PR #1242, opened 2024-01-19
fix(ssrServer): fastify scope resulted in hooks being called more than once

10xLaCroixDrinker:
## Description

Scopes the entire application within a top level plugin

## Motivation and Context

addresses fastify/fastify-cors#290

Currently rejected preflight requests result in an error because `preHandler` is called twice making it attempt to end timers that have already ended.

```
ERROR: Fastify application error: method OPTIONS, url "/success", correlationId "1fa8c448-db62-4552-968c-408857451b23", headersSent: false
    reqId: "req-1"
    error: {
      "type": "TypeError",
      "message": "The \"time\" argument must be an instance of Array. Received type number (9.911919)",
      "stack":
          TypeError [ERR_INVALID_ARG_TYPE]: The "time" argument must be an instance of Array. Received type number (9.911919)
              at process.hrtime (node:internal/process/per_thread:84:5)
              at endTimer (~/git/one-app-2/lib/server/utils/logging/fastifyPlugin.js:67:26)
              at Object.<anonymous> (~/git/one-app-2/lib/server/utils/logging/fastifyPlugin.js:189:5)
              at hookIterator (~/git/one-app-2/node_modules/fastify/lib/hooks.js:409:10)
              at next (~/git/one-app-2/node_modules/fastify/lib/hooks.js:243:18)
              at hookRunner (~/git/one-app-2/node_modules/fastify/lib/hooks.js:265:5)
              at notFound (~/git/one-app-2/node_modules/fastify/lib/reply.js:875:5)
              at Reply.callNotFound (~/git/one-app-2/node_modules/fastify/lib/reply.js:451:3)
              at Object.<anonymous> (~/git/one-app-2/node_modules/@fastify/cors/index.js:78:13)
              at preHandlerCallback (~/git/one-app-2/node_modules/fastify/lib/handleRequest.js:137:37)
      "code": "ERR_INVALID_ARG_TYPE"
    }
```

## How Has This Been Tested?

Ran the prod sample locally and sent the following request and validated the error no longer occurred:

```
$ curl --location --request OPTIONS 'http://localhost:3000/success' \
--header 'Origin: test.fail' \
--header 'Access-Control-Request-Method: GET' \
--header 'Access-Control-Request-Headers: content-type'
```

I also added logs to each of the handlers and tested the failing preflight before and after

**before**

Note that `preHandler` appears twice for `req-1`

```
WARN: onRequest 1
    reqId: "req-1"
WARN: onRequest 2
    reqId: "req-1"
[...]

PR #1243, opened 2024-01-19
fix(errorHandler): application errors would result in a 200 response

10xLaCroixDrinker:
## Description

Set HTTP status code to 500 in application error handler

## Motivation and Context

Application errors were causing HTTP 200 responses

## How Has This Been Tested?

Triggered an application error caused by bug fixed in #1242 validated that HTTP status of error page was 200 before change and 500 after

> Note: integration tests will fail until #1242 is merged

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

Correct HTTP status of application errors


Matthew-Mallimo:
Does the browser still display the content delivered when its a 500 response?

10xLaCroixDrinker:
> Does the browser still display the content delivered when its a 500 response?

Yes

dogpatch626 (approved):

dogpatch626 (approved):

code-forger (approved):

PR #1245, opened 2024-01-22
fix(devCdn): only cache 200's

code-forger:
## Description
Prevent the caching and re-playing of non 200 reponses.

## Motivation and Context
if there is a non-200 reponse, the cache will re-play the text content as a '200'. This means errors are stored in the cache and re-played incorrectly. 

Simply do not cache the non 200

## How Has This Been Tested?
TBT

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


code-forger:
> Ideally we would cache anything that is `ok == true` but then you would have to cache the status too.

Thats a better approach, will use that

smackfu:
Something to consider is that there is a single cache across the whole system, so any changes to the cache file have to be backwards compatible.  Which may make it impossible to cache status.

smackfu (approved):

bishnubista (approved):

10xLaCroixDrinker:
> > Ideally we would cache anything that is `ok == true` but then you would have to cache the status too.
> 
> Thats a better approach, will use that

@code-forger are you making this change?

code-forger:
@smackfu @10xLaCroixDrinker 

On second look, I think caching the status code is to much for this simple cache.

This is a CDN cache, the CDN should always be responding with 200's, I don't see any other 2XX code it would be useful to cache.

10xLaCroixDrinker (approved):

PR #1256, opened 2024-01-24
Feat/optional intl polyfill

timBarton96:
## Description
Added an environment variable to conditionally add the lean-intl polyfill.

## Motivation and Context
The lean-intl polyfill is missing Intl fetres and this PR is for consumers of one-app that want to trade compatibility with older browsers for these Intl features. The environment variable defaults to enable lean-intl polyfill, so needs to be explicitly be disabled. By disalbing the polyfill, the Intl implementations of the browsers are used and for modern browsers, this means extended functionality. 

## How Has This Been Tested?
I served a test-module which uses Luxon which uses Intl for its timezones functionality. The timezones would not work without the env var, but when the env var is enabled and there is no lean-intl the timezones work.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

Developers will now have the option to opt out of using lean-intl.


giulianok (changes requested):
there are too many unrelated and reverted changes, looks like the branch isn't up to date

timBarton96:
Closing this PR as it is out of sync with main and my attempts to fix have taken more effort than just opening a new branch and PR. New PR here: https://github.com/americanexpress/one-app/pull/1258

PR #1292, opened 2024-02-08
feat(logger): use a pretty logger in locale development with OTel enabled

10xLaCroixDrinker:
## Description

Default to pretty logger in console when using OTel for local development. Keep option to use OTel's console exporter when `--log-format=machine`

I also got rid of the extra logs that are printed during unit tests

## Motivation and Context

Make logs easier to consume during local dev when using OTel logging

## How Has This Been Tested?

Ran the application with the following combinations:

- otel: enabled, NODE_ENV=production
- otel: disabled, NODE_ENV=production
- otel: enabled, NODE_ENV=development
- otel: enabled, NODE_ENV=development, --log-format=machine
- otel: disabled, NODE_ENV=development
- otel: disabled, NODE_ENV=development, --log-format=machine

New and existing unit tests

## Types of Changes

- [x] Dependency update
- [ ] Security update

## Checklist:

More easily consumable logs by default during local dev when OTel is enabled


src/server/utils/logging/logger.js
```diff
@@ -26,24 +26,33 @@ import baseConfig from './config/base';
...
+    transportStreams.push(require('./config/development').default);
+  }
+
+  let transport;
+
+  if (transportStreams.length === 1) {
+    [transport] = transportStreams;
+  } else if (transportStreams.length > 1) {
+    transport = multistream(transportStreams);
   }
```
giulianok, line 53: this could be converted into a small fn to make it more modular and readable

```suggestion
const getTransport = (transportStreams = []) => {
   if (transportStreams.length === 1) return transportStreams[0];
   if (transportStreams.length > 1) return multistream(transportStreams);
}

...

const transport = getTransport(transportStreams);
```

src/server/utils/logging/logger.js
```diff
@@ -26,24 +26,33 @@ import baseConfig from './config/base';
...
-  if (!useProductionConfig && !process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT) {
+  if (!useProductionConfig) {
     // eslint-disable-next-line global-require -- do not load development logger in production
-    transport = require('./config/development').default;
-  } else if (process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT) {
-    transport = createOtelTransport();
+    transportStreams.push(require('./config/development').default);
+  }
+
+  let transport;
```
giulianok, line 47: what if `transportStreams` is an empty array?

src/server/utils/logging/logger.js
```diff
@@ -26,24 +26,33 @@ import baseConfig from './config/base';
...
+  }
 
-  if (!useProductionConfig && !process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT) {
+  if (!useProductionConfig) {
     // eslint-disable-next-line global-require -- do not load development logger in production
-    transport = require('./config/development').default;
-  } else if (process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT) {
-    transport = createOtelTransport();
+    transportStreams.push(require('./config/development').default);
+  }
```
giulianok, line 45: the problem with mutation, especially here, is that cases for transport streams and for pino config are different, I would convert them into functions to make it more readable, like

```js
const getTransportStreams = () => {
  const transportStreams = [];

  if (process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT) {
    transportStreams.push(createOtelTransport({
      grpc: true,
      console: process.env.NODE_ENV === 'development' && argv.logFormat === 'machine',
    }));
  }

  if (!useProductionConfig) {
    // eslint-disable-next-line global-require -- do not load development logger in production
    transportStreams.push(require('./config/development').default);
  }

  return transportStreams;
}

const getPinoConfig = () => {
  if (process.env.OTEL_EXPORTER_OTLP_LOGS_ENDPOINT) {
    return deepmerge(baseConfig, otelConfig);
  }

  if (useProductionConfig) {
    return deepmerge(baseConfig, productionConfig);
  }

  return baseConfig
}
```

giulianok (changes requested):

src/server/utils/logging/logger.js
10xLaCroixDrinker: then we don't have a transport. it is not required

10xLaCroixDrinker: if `transportStreams` is an empty array the variable is not used, and we passed undefined for the transport

src/server/utils/logging/config/otel.js
```diff
@@ -53,16 +43,24 @@ export function createOtelTransport() {
...
+  if (consoleExporter) {
+    logRecordProcessorOptions.push({
       recordProcessorType: 'simple',
       exporterOptions: { protocol: 'console' },
-    }];
+    });
+  }
+
+  if (logRecordProcessorOptions.length === 1) {
+    [logRecordProcessorOptions] = logRecordProcessorOptions;
```
JAdshead, line 63: is this intended for `logRecordProcessorOptions` to either be empty array or array of objects or single object`{}` ?

10xLaCroixDrinker: Yes

JAdshead (approved):

giulianok (approved):

PR #1302, opened 2024-02-22
refactor(deps): remove ip for node native implementation

dogpatch626:
## Description

remove node package 'Ip' use node native implementation in order to provide the same functionality 
## How Has This Been Tested?

unit tests, built local docker image

## Types of Changes

- [x] Dependency update
- [ ] Security update

## Checklist:


src/server/utils/getIP.js
```diff
@@ -0,0 +1,12 @@
+import { networkInterfaces } from 'os';
+
+export const getIp = () => {
+  const interfaces = networkInterfaces();
+
+  const ipAddress = Object.keys(interfaces)
```
giulianok, line 6:
```suggestion
  const ipAddresses = Object.keys(interfaces)
```

just because it's an array

src/server/utils/getIP.js
```diff
@@ -0,0 +1,12 @@
...
+    .map((name) => interfaces[name].find((iface) => iface.family === 'IPv4'
+            && !iface.internal && iface.address !== '127.0.0.1'))
+    .filter(Boolean);
+
+  return ipAddress[0] ? ipAddress[0].address : '0.0.0.0';
```
giulianok, line 11: the `ip` package actually uses `os`, it just provides more utilities, not sure if it would be better to remove it or just keep it in case we need some of the utilities in the future

giulianok (changes requested):

src/server/config/env/runTime.js
```diff
@@ -252,5 +252,5 @@ const runTime = [
   },
 ];
 runTime.forEach(preprocessEnvVar);
-export { ip };
```
giulianok, line 255: `ip` is actually a `string` but we return a function, should we return the function call?

giulianok (changes requested):

src/server/utils/getIP.js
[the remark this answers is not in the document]
dogpatch626: @Matthew-Mallimo

Matthew-Mallimo (approved):

giulianok (approved):

PR #1328, opened 2024-03-14
feat: set environment variables from secrets file

Matthew-Mallimo:
## Description

Adds the ability to load secrets from file (path set by a new env var `ONE_ENVIRONMENT_SECRETS_PATH`) and expose them as environment variables.

## Motivation and Context

We want to allow users to load secrets from a file and expose them as environment variables.

## How Has This Been Tested?

Unit, running in an environment internally.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:


10xLaCroixDrinker:
This could cause some unexpected behavior if the user tries to load OTEL vars in this file. You may need to conditionally add a `--require=dotenv/config` to `start.sh`

PR #1343, opened 2024-03-26
fix(server): devCDN should not crash when bundles are not present

PixnBits:
## Description

The development CDN crashes when performing the substitution in the URL of a nonexistant bundle.

## Motivation and Context

The legacy browser bundle is optional, not required.

```
err: {
    "type": "TypeError",
    "message": "Cannot read properties of undefined (reading 'url')",
    "stack":
        TypeError: Cannot read properties of undefined (reading 'url')
            at .../one-app/lib/server/utils/devCdnFactory.js:44:53
```

## How Has This Been Tested?

Unit tests added that have only one bundle each.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

Resolve the server crashing on them when running with modules that have not opted in to building the legacy browser bundles.

### DIFF
--- a/src/server/utils/devCdnFactory.js
+++ b/src/server/utils/devCdnFactory.js
@@ -33,9 +33,15 @@ const getLocalModuleMap = ({ pathToModuleMap, oneAppDevCdnAddress }) => {
   const moduleMap = JSON.parse(fs.readFileSync(pathToModuleMap, 'utf8').toString());
   Object.keys(moduleMap.modules).forEach((moduleName) => {
     const module = moduleMap.modules[moduleName];
-    module.browser.url = module.browser.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
-    module.legacyBrowser.url = module.legacyBrowser.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
-    module.node.url = module.node.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
+    Object
+      .values(module)
+      .filter((bundle) => Object.hasOwnProperty.call(bundle, 'url'))
+      .forEach((bundle) => {
+        /* eslint-disable-next-line no-param-reassign -- the in-memory copy is created here during
+        // the read from disk, so the replacement side-effect of this loop is local to the
+        // getLocalModuleMap function and not to any arguments */
+        bundle.url = bundle.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
+      });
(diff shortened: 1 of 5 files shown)


src/server/utils/devCdnFactory.js
```diff
@@ -33,9 +33,15 @@ const getLocalModuleMap = ({ pathToModuleMap, oneAppDevCdnAddress }) => {
   const moduleMap = JSON.parse(fs.readFileSync(pathToModuleMap, 'utf8').toString());
   Object.keys(moduleMap.modules).forEach((moduleName) => {
     const module = moduleMap.modules[moduleName];
-    module.browser.url = module.browser.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
-    module.legacyBrowser.url = module.legacyBrowser.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
-    module.node.url = module.node.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
+    Object
+      .values(module)
+      .filter((bundle) => Object.hasOwnProperty.call(bundle, 'url'))
```
PixnBits, line 38: this is defensive; do we need to be? e.g. is `baseUrl` always computed or can it be part of the module map JSON resource?

code-forger (approved):

JAdshead (dismissed):

code-forger (changes requested):
This Pr fixes a symptom, not a cause, the root cause is fixed elsewhere

src/server/utils/devCdnFactory.js
```diff
@@ -72,15 +78,15 @@ const consumeRemoteRequest = async (remoteModuleMapUrl, hostAddress, remoteModul
...
-      );
+      Object
+        .values(module)
+        .filter((bundle) => Object.hasOwnProperty.call(bundle, 'url'))
+        .forEach((bundle) => {
+          /* eslint-disable-next-line no-param-reassign -- the in-memory copy is created here during
+          // the read from network, so the replacement side-effect of this loop is local to the
+          // consumeRemoteRequest function and not to any arguments */
+          bundle.url = bundle.url.replace(new URL(bundle.url).origin, oneAppDevStaticsAddress);
+        });
```
PixnBits, line 89: I'm not sure where the test(s) for the previous implementation is/are, so this needs attention

PixnBits:
> This Pr fixes a symptom, not a cause, the root cause is fixed elsewhere

https://github.com/americanexpress/one-app-cli/pull/629

PR #1368, opened 2024-04-04
Fix: watch local modules other bundles, 5.x.x

PixnBits:
## Description

During development use filesystem polling to detect modules being built and then reload them in the running server. This increases the filesystem and CPU usage, but is much more reliable.

This also adds some log entries for the user to know when a module was reloaded, removing guesswork.

Some unnecessary disk reads (module map) were removed.

## Motivation and Context

Users are hitting development scenarios where new builds of modules are not detected and then not reloaded in the running server. The [limitations of the Node.js built-in API](https://nodejs.org/docs/latest-v18.x/api/fs.html#fswatchfilename-options-listener) are important, but are smoothed over by [chokidar](https://www.npmjs.com/package/chokidar). However, even chokidar is not accurately emiting events, sometimes due to superseding situations like the modules being docker volume mounts, e.g. https://forums.docker.com/t/file-system-watch-does-not-work-with-mounted-volumes/12038.

## How Has This Been Tested?

I've both 
* `$ npm serve-module some-module` and run the server
* built the image and provided the digest to [@americanexpress/one-app-runner](https://github.com/americanexpress/one-app-cli/tree/main/packages/one-app-runner) in a module 

Before these changes running on metal would occasionally miss the file writing finishing and not reload, and running in a docker container every change after the first build would be missed. After the changes in this PR I saw every write finish detected.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

- [x] These changes should be applied to a maintenance branch. (This is that PR, see #1370)

Increased confidence when changes are loaded, and end (or at the very least, greatly decrease) the users having to restart the one-app server to see server bundle changes loaded.

### DIFF
--- a/__tests__/server/utils/watchLocalModules.spec.js
+++ b/__tests__/server/utils/watchLocalModules.spec.js
@@ -26,7 +26,7 @@ import { getModules, resetModuleRegistry } from 'holocron/moduleRegistry';
 import watchLocalModules from '../../../src/server/utils/watchLocalModules';
 
 jest.mock('fs', () => {
-  const fs = jest.requireActual('fs');
+  const fsActual = jest.requireActual('fs');
   const setImmediateNative = global.setImmediate;
 
   let mockedFilesystem;
@@ -54,18 +54,19 @@ jest.mock('fs', () => {
   } */
 
   function getEntry(parts) {
-    let current = mockedFilesystem;
-    for (const entryName of parts) {
-      if (!current || !current.has('entries')) {
+    return parts.reduce((parentEntry, entryName) => {
+      if (!parentEntry || !parentEntry.has('entries')) {
         return null;
       }
-      current = current.get('entries').get(entryName);
-    }
-
-    return current;
+      return parentEntry.get('entries').get(entryName);
+    }, mockedFilesystem);
   }
 
   let inodeCount = 0;
+  function getNewInode() {
+    inodeCount += 1;
+    return inodeCount;
+  }
   const mock = {
     clear() {
       const createdMillis = Date.now() + Math.floor(Math.random() * 1e4) / 1e4;
@@ -81,7 +82,7 @@ jest.mock('fs', () => {
             gid: 512,
             rdev: 0,
             blksize: 4096,
-            ino: ++inodeCount,
+            ino: getNewInode(),
             size: 128,
             blocks: 0,
             atimeMs: Date.now() + Math.floor(Math.random() * 1e4) / 1e4,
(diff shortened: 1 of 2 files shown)


src/server/utils/watchLocalModules.js
```diff
@@ -17,54 +17,165 @@
 // This file is only used in development so importing devDeps is not an issue
 /* eslint "import/no-extraneous-dependencies": ["error", {"devDependencies": true}] */
 
-import fs from 'fs';
-import path from 'path';
-import chokidar from 'chokidar';
```
PixnBits, line 22: if this is the only place chokidar is used (directly), can we remove it from package.json?

__tests__/server/utils/watchLocalModules.spec.js
```diff
@@ -59,190 +307,1028 @@ jest.mock('holocron/moduleRegistry', () => {
...
+    expect(setImmediate).toHaveBeenCalledTimes(3);
+    expect(setTimeout).toHaveBeenCalledTimes(6);
+
+    expect(loadModule).toHaveBeenCalled();
   });
 
-  it('should ignore changes to all files but the server bundle', async () => {
+  // instance when the CHANGE_WATCHER_INTERVAL and WRITING_FINISH_WATCHER_TIMEOUT lined up
+  // we need to avoid scheduling the write watcher like a tree (many branches, eventually eating all CPU and memory)
+  it.todo('should schedule watching for writes only once when both watchers have run');
```
PixnBits, line 1036: TODO (hence the draft PR)

src/server/utils/watchLocalModules.js
```diff
@@ -17,54 +17,165 @@
...
+            return fileStat;
+          })
+      );
+    }
 
-  watcher.on('change', async (changedPath) => {
-    try {
-      if (!changedPath.endsWith('.node.js')) return;
+    return stating.get(filePath);
+  }
```
PixnBits, line 105: too much? (remove?)

PixnBits: removed in [7398793](https://github.com/americanexpress/one-app/pull/1368/commits/73987933a31da700d91eb06fef6a359b35da3220)

__tests__/server/utils/watchLocalModules.spec.js
PixnBits: did/done

PixnBits:
```
> commitlint --to HEAD --from origin/5.x.x

⧗   input: test(server): use node test environment, timers not working\ninconsistentcies in the event loop evaluation, so switched to manual timer mocks
✖   header must not be longer than 100 characters, current length is 141 [header-max-length]

✖   found 1 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint

⧗   input: fix(server): more reliable module bundle update watching
⚠   body must have leading blank line [body-leading-blank]
⚠   body's lines must not be longer than [10](https://github.com/americanexpress/one-app/actions/runs/8559772898/job/23457261069?pr=1368#step:9:11)0 characters [body-max-line-length]
```
https://github.com/americanexpress/one-app/actions/runs/8559772898/job/23457261069?pr=1368

editing via rebase

PR #1370, opened 2024-04-05
Fix: watch local modules other bundles

PixnBits:
## Description

During development use filesystem polling to detect modules being built and then reload them in the running server. This increases the filesystem and CPU usage, but is much more reliable.

This also adds some log entries for the user to know when a module was reloaded, removing guesswork.

Some unnecessary disk reads (module map) were removed.

## Motivation and Context

Users are hitting development scenarios where new builds of modules are not detected and then not reloaded in the running server. The [limitations of the Node.js built-in API](https://nodejs.org/docs/latest-v18.x/api/fs.html#fswatchfilename-options-listener) are important, but are smoothed over by [chokidar](https://www.npmjs.com/package/chokidar). However, even chokidar is not accurately emiting events, sometimes due to superseding situations like the modules being docker volume mounts, e.g. https://forums.docker.com/t/file-system-watch-does-not-work-with-mounted-volumes/12038.

## How Has This Been Tested?

I've both 
* `$ npm serve-module some-module` and run the server
* built the image and provided the digest to [@americanexpress/one-app-runner](https://github.com/americanexpress/one-app-cli/tree/main/packages/one-app-runner) in a module 

Before these changes running on metal would occasionally miss the file writing finishing and not reload, and running in a docker container every change after the first build would be missed. After the changes in this PR I saw every write finish detected.

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

- [x] These changes should be applied to a maintenance branch. (#1368)

Increased confidence when changes are loaded, and end (or at the very least, greatly decrease) the users having to restart the one-app server to see server bundle changes loaded.


src/server/utils/watchLocalModules.js
```diff
@@ -14,52 +14,145 @@
...
+  if (!moduleMapEntry) {
+    console.warn(`module "${moduleName}" not in the module map, make sure to serve-module first`);
+    return;
+  }
+
+  console.log(`the Node.js bundle for ${moduleName} finished saving, attempting to load`);
+
+  let newModule;
+  try {
+    // FIXME: this leads to a race condition later
```
code-forger, line 62: I'm not sure there is a race condition

You get, edit, and reset the module map below without any await, so yes, the modules will race to load. But regardless the edits to the module map will be atomic?

code-forger (commented):
See comment

src/server/utils/watchLocalModules.js
```diff
@@ -14,52 +14,145 @@
...
+  const [moduleName] = changedPath.split(path.sep);
+  const moduleMap = getModuleMap();
+  const moduleMapEntry = moduleMap.getIn(['modules', moduleName]);
+
+  if (!moduleMapEntry) {
+    console.warn(`module "${moduleName}" not in the module map, make sure to serve-module first`);
+    return;
+  }
+
+  console.log(`the Node.js bundle for ${moduleName} finished saving, attempting to load`);
```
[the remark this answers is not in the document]
PixnBits: depends, if the module takes a long time to load it would be useful to the user to know that the process has started

src/server/utils/watchLocalModules.js
PixnBits: different copies are created if two modules are loaded at the same time

src/server/utils/watchLocalModules.js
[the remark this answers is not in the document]
PixnBits: changed the levels in d32878d

src/server/utils/watchLocalModules.js
```diff
@@ -14,52 +14,145 @@
...
   getModules,
   getModuleMap,
   resetModuleRegistry,
   addHigherOrderComponent,
 } from 'holocron/moduleRegistry';
-import { getIp } from './getIP';
 import onModuleLoad from './onModuleLoad';
 
+const CHANGE_WATCHER_INTERVAL = 1000;
+const WRITING_FINISH_WATCHER_TIMEOUT = 400;
```
PixnBits, line 30: these should be configurable by the user
I'm debating via [arguments](https://github.com/americanexpress/one-app/blob/main/src/server/config/env/argv.js) or [environment variables](https://github.com/americanexpress/one-app/blob/main/src/server/config/env/runTime.js)

PixnBits: perhaps we let that be another PR so as to get this fix in, and then resolve the plan?

10xLaCroixDrinker: I'd prefer arguments. Particularly since this is just local dev. They are much more discoverable since they are self documenting. Also those using one-app-runner will be able to very easily include them in the runner config.

10xLaCroixDrinker (commented):
What do you think of holding this until Node.js 20 is in, and using [`fs.watch`](https://nodejs.org/docs/latest-v20.x/api/fs.html#fswatchfilename-options-listener)? It should simplify a lot of this.

There are some [caveats](https://nodejs.org/docs/latest-v20.x/api/fs.html#caveats) to it though.

> On Windows, no events will be emitted if the watched directory is moved or renamed. An EPERM error is reported when the watched directory is deleted.

I think this one will be fine.

> On IBM i systems, this feature is not supported.

I think this one might be inaccurate, based on the following release note for `fs.watch`

| Version | Changes |
|:---|:---|
| v19.1.0 | Added recursive support for Linux, AIX and IBMi. |

> On Linux and macOS systems, fs.watch() resolves the path to an inode and watches the inode. If the watched path is deleted and recreated, it is assigned a new inode. The watch will emit an event for the delete but will continue watching the original inode. Events for the new inode will not be emitted. This is expected behavior.

Sounds like if we run into this issue, we may be able to listen for the delete and recreate the watcher.

src/server/utils/watchLocalModules.js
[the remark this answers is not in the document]
code-forger: Sorry, I didn't mean for you to make both of these `info`

The second log, that tells the user something _has_ been reloaded should be `log` log level still.

This is an important piece of information for the deverloper, and the lack of this notification is currently a point of frustration for Devs.

That said, I really like your time based de-duplicattion here. If all is well, they see 1 log saying success, if it's taking way longer than normal, they see a 'attempting to reload' message, if they then never get the success message they know something is wrong.

In _that_ case, both of these can be `log` log level.

We just have to be sure that 500ms is the right time budget

PR #1372, opened 2024-04-08
DX enhancements - quieter logging on startup and polling, clickable URL in console

agorskiy12:
## Description
when running one-app in node.env = 'development'

logging is cleaner on startup
hiding messages about bundle sizes unless they are critical
hiding messages about what ports the app is launching on
hiding messages about module map polling, unless something goes wrong
offer a clickable URL to the UI project in the command line

## Motivation and Context
GitHub discussion about DX enhancements 

## How Has This Been Tested?
unit test

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

they will see less logging in development


src/server/utils/cdnCache.js
```diff
@@ -26,15 +26,18 @@ export const oneAppModuleCachePath = path.join(oneAppDirectoryPath, cacheFileNam
...
-    console.log(chalk.bold.cyanBright('CACHE INFORMATION'));
-    console.log(message);
-    console.log('To clear the cache, please delete this file:');
-    console.log(`    ${chalk.bold.cyanBright(path.join('~', oneAppDirectoryName, cacheFileName))}`);
-    console.log(chalk.bold.cyanBright(separator));
+    console.info(chalk.bold.cyanBright(separator));
+    console.info(chalk.bold.cyanBright('CACHE INFORMATION'));
+    console.info(message);
+    console.info('To clear the cache, please delete this file:');
+    console.info(`    ${chalk.bold.cyanBright(path.join('~', oneAppDirectoryName, cacheFileName))}`);
```
10xLaCroixDrinker, line 39: Can you replace the template literal with string interpolation? Doing so helps with performance in cases where the log is dropped and does not have to be evaluated. (Same applies to `message` a couple lines above)

src/server/utils/cdnCache.js
```diff
@@ -26,15 +26,18 @@ export const oneAppModuleCachePath = path.join(oneAppDirectoryPath, cacheFileNam
 export const showCacheInfo = async () => {
   try {
     const stats = await fsPromises.stat(oneAppModuleCachePath);
+    if (stats.size > 1024 * 1024 * 10) {
+      console.warn(chalk.bold.redBright('🚨🚨🚨: Cache file size is more than 10MB. 🚨🚨🚨'));
```
10xLaCroixDrinker, line 30: Isn't this pretty common for large tenancies?

10xLaCroixDrinker (changes requested):

src/server/utils/cdnCache.js
code-forger: Yah I don't think this warning is necessary, or should be up at 300mb

src/server/utils/cdnCache.js
agorskiy12: do you mean replace with string concatenation?

src/server/utils/cdnCache.js
agorskiy12: i guess ill just delete it.

src/server/utils/cdnCache.js
smackfu: Like:
```
console.info('one-app-dev-cdn loading module map from %s', remoteModuleMapUrl);
```

src/server/utils/cdnCache.js
code-forger: @smackfu the cache only keeps one copy of any file for a module. It doesn't not keep multiple versions. So if you only work on one tenancy, your cache size is pretty 'stable' because new versions of your modules replace old versions

src/server/utils/pollModuleMap.js
```diff
@@ -126,7 +126,7 @@ async function pollModuleMap() {
   recordPollingForMonitor();
   startPollingMonitorIfNotAlready();
   try {
-    console.log('pollModuleMap: polling...');
+    console.info('pollModuleMap: polling...');
```
PixnBits, line 129: lots of log level changes
how does this affect the log entries in production? are there some side effects we and users should be aware of? are they breaking? (e.g. searching through logs)

10xLaCroixDrinker: Info is the default log level in production

src/server/utils/cdnCache.js
10xLaCroixDrinker: I agree with Matt, we could eliminate a lot of this.

src/server/utils/cdnCache.js
```diff
@@ -26,15 +26,14 @@ export const oneAppModuleCachePath = path.join(oneAppDirectoryPath, cacheFileNam
...
-    console.log(`    ${chalk.bold.cyanBright(path.join('~', oneAppDirectoryName, cacheFileName))}`);
-    console.log(chalk.bold.cyanBright(separator));
+    const fileSizeInMB = stats.size / (1024 * 1024); // bytes to mb
+    const separator = '*'.repeat(50);
+    console.info(chalk.bold.cyanBright(separator));
+    console.info(chalk.bold.cyanBright('CACHE INFORMATION'));
+    console.info('File size of %s: %s', cacheFileName, chalk.bold.greenBright(fileSizeInMB.toFixed(2), 'MB'));
+    console.info('To clear the cache, please delete this file:');
+    console.info('    %s', chalk.bold.redBright(path.join('~', oneAppDirectoryName, cacheFileName)));
+    console.info(chalk.bold.cyanBright(separator));
```
10xLaCroixDrinker, line 36:
```suggestion
    const fileSizeInMB = stats.size / (1024 * 1024); // bytes to mb
    const cachePath = path.join('~', oneAppDirectoryName, cacheFileName);
    console.info('Local module cache size is %sMB. To clear the cache, delete %s', fileSizeInMB.toFixed(2), cachePath);
```

src/server/index.js
```diff
@@ -57,7 +57,10 @@ export const listen = async ({
       port,
     });
 
-    console.log('%s listening on port %d', context, port);
+    console.info('%s listening on port %d', context, port);
+    if (context === '🌎 One App server') {
+      console.log('🚀 One App Server is running on http://localhost:%d 🚀', port);
```
10xLaCroixDrinker, line 62: This URL is only accurate in development, but will be logged in production

10xLaCroixDrinker (changes requested):

src/server/utils/cdnCache.js
```diff
@@ -26,15 +25,9 @@ export const oneAppModuleCachePath = path.join(oneAppDirectoryPath, cacheFileNam
...
-    const separator = '*'.repeat(message.length);
-    console.log(chalk.bold.cyanBright(separator));
-    console.log(chalk.bold.cyanBright('CACHE INFORMATION'));
-    console.log(message);
-    console.log('To clear the cache, please delete this file:');
-    console.log(`    ${chalk.bold.cyanBright(path.join('~', oneAppDirectoryName, cacheFileName))}`);
-    console.log(chalk.bold.cyanBright(separator));
+    const fileSizeInMB = stats.size / (1024 * 1024); // bytes to mb
+    const cachePath = path.join('~', oneAppDirectoryName, cacheFileName);
+    console.info('Local module cache size is %sMB. To clear the cache, delete %s', fileSizeInMB.toFixed(2), cachePath);
```
dogpatch626, line 30: should this be info? I feel like as this is a development only addition it makes sense to keep this logged out during the beginning especially as it is only logged once.

dogpatch626 (approved):

src/server/index.js
10xLaCroixDrinker: I think we should actually add a new log level for cases like this. `console.dev`

It would require a change here:

https://github.com/pinojs/pino-abstract-transport/pull/90/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346R67-R71

and here

https://github.com/americanexpress/one-app/blob/main/src/server/utils/logging/monkeyPatchConsole.js#L19

```
  customLevels: {
    dev: process.env.NODE_ENV === 'development' ? 34 : 0,
    log: 35,
  },
```

This would allow us to have logs that are only for development, but not debugging, without carrying around this condition

10xLaCroixDrinker: For reference in the above, `console.info` is `30` and `console.warn` is `40`

10xLaCroixDrinker: @giulianok that kind of defeats the purpose of the level. the existing log levels (trace, debug, info, log, warn, error, fatal) all have practical cases in both development and productions. these would strictly be for development.

If the point is about using console though, @PixnBits also pointed out and I agreed we should replace out use of `console[method]` with importing the logger and using `logger[method]`. IIRC everything left in src that uses `console[method]` is outside the fastify context. Anything within that context should be using `fastify.log[method]` or `req.log[method]` (if in the context of a request)

src/server/utils/cdnCache.js
agorskiy12: this one could be a good candidate for the new logger.dev logging method

src/server/utils/logging/config/base.js
```diff
@@ -21,6 +21,7 @@ export default {
   level: argv.logLevel,
   customLevels: {
     log: 35,
+    dev: 36,
```
10xLaCroixDrinker, line 24:
```suggestion
    dev: process.env.NODE_ENV === 'development' ? 34 : 0,
    log: 35,
```

src/server/utils/logging/config/development.js
```diff
@@ -29,11 +29,12 @@ export const pinoPrettyOptions = {
     debug: 20,
     info: 30,
     log: 35,
+    dev: 36,
     warn: 40,
     error: 50,
     fatal: 60,
   },
-  customColors: 'trace:white,debug:green,info:gray,log:blue,warn:yellow,error:red,fatal:bgRed',
+  customColors: 'trace:white,debug:green,info:gray,log:blue,dev:green,warn:yellow,error:red,fatal:bgRed',
```
10xLaCroixDrinker, line 37: Is there a color we can use that we're not already using?

Maybe we _should_ use magenta.

agorskiy12: i like magenta. but where do I find the list of accepted colors? I was trying with "orange" or "purple" but those didnt work

src/server/utils/cdnCache.js
```diff
@@ -26,15 +26,9 @@ export const oneAppModuleCachePath = path.join(oneAppDirectoryPath, cacheFileNam
...
-    console.log(chalk.bold.cyanBright('CACHE INFORMATION'));
-    console.log(message);
-    console.log('To clear the cache, please delete this file:');
-    console.log(`    ${chalk.bold.cyanBright(path.join('~', oneAppDirectoryName, cacheFileName))}`);
-    console.log(chalk.bold.cyanBright(separator));
+    const fileSizeInMB = stats.size / (1024 * 1024); // bytes to mb
+    const cachePath = path.join('~', oneAppDirectoryName, cacheFileName);
+    logger.dev('Local module cache size is %sMB. To clear the cache, delete %s', fileSizeInMB.toFixed(2), cachePath);
   } catch (error) {
     console.error('There was error checking file stat', error);
```
10xLaCroixDrinker, line 33: Don't forget to changes the other instances of console to logger

PR #1375, opened 2024-04-11
feat(Dockerfile): use .nvmrc to build the image (5.x.x)

PixnBits:
## Description

use .nvmrc to build the image
and run workflows, creating a single point to define the Node.js version to use and thus making upgrades easier

## Motivation and Context

Avoid the myriad of locations where Node.js versions are recorded, i.e. Dockerfile and workflows. package.json engines will still need manual updates though (perhaps that can be a workflow, or a unit test to ensure it is satisfied by the contents of `.nvmrc`?)

## How Has This Been Tested?

built locally, though not run, but did see the expected label used when manually running `$ docker build . --build-arg VERSION=18.17.0` and `docker build . --build-arg VERSION=$(cat .nvmrc)`

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

- [x] These changes should be applied to a maintenance branch. (this is that PR, see #1090)

faster Node.js version updates? this is more for the maintainers so no impact to devs is the basic requirement 😅


PixnBits:
should we change https://github.com/americanexpress/one-app/blob/24dca2de9d97db021143500c868e078ea7b1eb6d/package.json#L6-L7 to include 16.x 
or https://github.com/americanexpress/one-app/blob/24dca2de9d97db021143500c868e078ea7b1eb6d/.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml#L13-L14 to remove 16.x ?

10xLaCroixDrinker:
> should we change
> 
> https://github.com/americanexpress/one-app/blob/24dca2de9d97db021143500c868e078ea7b1eb6d/package.json#L6-L7
> 
> to include 16.x
> or
> https://github.com/americanexpress/one-app/blob/24dca2de9d97db021143500c868e078ea7b1eb6d/.github/workflows/on-pr-and-trunk_one-app-unit-and-lint-tests.yml#L13-L14
> 
> to remove 16.x ?

Remove 16

package.json
```diff
@@ -4,8 +4,8 @@
   "description": "One Amex SPA technology stack.",
   "main": "index.js",
   "engines": {
-    "node": ">=12",
-    "npm": ">=6"
+    "node": "^18",
+    "npm": "^9"
```
10xLaCroixDrinker, line 8: Your engines change is causing all workflows that run install to fail.

10xLaCroixDrinker (changes requested):

code-forger (approved):

PR #1382, opened 2024-04-11
fix(server): remove npm and yarn from the production image

PixnBits:
## Description

Remove npm, npx, yarn, corepack, and their `node_modules` directories from the production image artifact.

### Upsides
* shipping less to production

### Why we should not do this
* `npm serve-module <module-path>` will not work; not an issue in production but will be an issue for anyone using the production image with [@americanexpress/one-app-runner](https://github.com/americanexpress/one-app-cli/tree/main/packages/one-app-runner)
* ~~removing files adds to the image size: similar to a git diff, the difference is stored.~~ this must have changed, measuring this on my machine with [dive](https://github.com/wagoodman/dive) and [docker history](https://docs.docker.com/reference/cli/docker/image/history/) I see the layer size being 0 B

## Motivation and Context

Ship as little as possible to production to minimize attack surface area (unlikely to contribute to an exploit, but computers are complex enough to make that risk not zero).

## How Has This Been Tested?

PR checks

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

serving modules with npm (as @americanexpress/one-app-runner does) would not work anymore (explained above)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,9 @@ ENV NODE_ENV=production
 # see src/server/config/env/runtime.js
 EXPOSE 3000
 EXPOSE 3005
+USER root
+RUN rm -Rf /usr/local/lib/node_modules /opt/yarn-* && \
+    rm /usr/local/bin/corepack /usr/local/bin/npm /usr/local/bin/npx /usr/local/bin/yarn /usr/local/bin/yarnpkg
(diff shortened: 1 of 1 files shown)


10xLaCroixDrinker:
Is this a common practice?

PR #1387, opened 2024-04-13
fix(server): handle kernel signals by including tini

PixnBits:
## Description

Wrap the Node.js process with the tini init wrapper. This enables proper kernel signal handling inside the container, removing the need for forceful shutdowns (e.g. `$ docker ps; docker kill <container-id>`).

This adds 96kB to the image size.

We may later be able to remove [src/server/shutdown.js](https://github.com/americanexpress/one-app/blob/v6.11.0/src/server/shutdown.js) if the Babel issue is resolved (it's been a number of years). Though, it might serve has a failure handling in the event we miss something that keeps Node.js running (e.g. a `setInterval` without `.unref()`).

## Motivation and Context

> Node.js was not designed to run as PID 1 which leads to unexpected behaviour when running inside of Docker. For example, a Node.js process running as PID 1 will not respond to SIGINT (CTRL-C) and similar signals. ...
> You can...include Tini [directly in your Dockerfile](https://github.com/krallin/tini#using-tini), ensuring your process is always started with an init wrapper.

https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals

## How Has This Been Tested?

I've built the image locally, used it via @americanexpress/one-app-runner (`npm start -- --offline --docker-image=sha256:...`) and upon <kbd>CTRL</kbd>+<kbd>C</kbd> the container stops (due to the Node.js process ending).

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

- [x] These changes should be applied to a maintenance branch. (#1388)

Don't have to wait 10,000ms for the shutdown check to run. Should also be more graceful and faster shutdowns in production.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,8 @@ RUN NODE_ENV=production npm run build && \
 # development image
 # docker build . --target=development
 FROM node:$VERSION-alpine as development
+# https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals
+RUN apk add --no-cache tini
 ARG USER
 ENV USER ${USER:-node}
 ENV NODE_ENV=development
@@ -50,6 +52,8 @@ COPY --from=builder --chown=node:node /opt/one-app/development ./
 # production image
 # last so that it's the default image artifact
 FROM node:$VERSION-alpine as production
+# https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals
+RUN apk add --no-cache tini
 ARG USER
 ENV USER ${USER:-node}
 ENV NODE_ENV=production
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,11 +29,14 @@ RUN NODE_ENV=production npm run build && \
     mv /opt/build/bundle.integrity.manifest.json /opt/one-app/production && \
     mv /opt/build/.build-meta.json /opt/one-app/production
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -32,6 +32,8 @@ RUN NODE_ENV=production npm run build && \
 # development image
 # docker build . --target=development
 FROM node:$VERSION-alpine as development
+# https://github.com/nodejs/docker-node/blob/main/docs/BestPractices.md#handling-kernel-signals
+RUN apk add --no-cache tini
```
PixnBits, line 36: maybe we can run this once and create a "wrapped" build stage to use as the base for development and production targets

10xLaCroixDrinker (changes requested):
This breaks the start command when running in the repo.

```sh
$ npm start

> @americanexpress/one-app@6.12.1 start
> scripts/start.sh

/sbin/tini -- node --dns-result-order ipv4first --no-experimental-fetch --require=dotenv/config lib/server/index.js
scripts/start.sh: line 39: /sbin/tini: No such file or directory
scripts/start.sh: line 39: exec: /sbin/tini: cannot execute: No such file or directory
```

PixnBits:
good catch! it's also probably better for tini to be the entrypoint rather than deferred to a script file so it can manage all the child processes
updated in fe135a

Dockerfile
PixnBits: updated in fe13e5a

code-forger (approved):

PR #1391, opened 2024-04-18
fix(watchLocalModules): fix missing baseUrl

code-forger:
## Description
This fixes users seeing the following log when re-building locally served modules running the dev server:
```Encountered error fetching module at undefinedmodule-config.json: Only absolute URLs are supported```

## Motivation and Context
This bug means the module-config.json will never load, and the users sees an error they have no way of understanding

## How Has This Been Tested?
running agains reproduction case
unit tests

## Types of Changes

- [ ] Dependency update
- [ ] Security update

## Checklist:

No more confusing log

### DIFF
--- a/src/server/utils/watchLocalModules.js
+++ b/src/server/utils/watchLocalModules.js
@@ -26,6 +26,7 @@ import {
 } from 'holocron/moduleRegistry';
 import { getIp } from './getIP';
 import onModuleLoad from './onModuleLoad';
+import addBaseUrlToModuleMap from './addBaseUrlToModuleMap';
 
 export default function watchLocalModules() {
   const staticsDirectoryPath = path.resolve(__dirname, '../../../static');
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/server/utils/watchLocalModules.js [hunk 1]
```diff
@@ -43,12 +44,13 @@ export default function watchLocalModules() {
...
       const moduleMap = JSON.parse(fs.readFileSync(moduleMapPath, 'utf8'));
 
-      const moduleData = moduleMap.modules[moduleNameChangeDetectedIn];
+      const moduleData = addBaseUrlToModuleMap(moduleMap).modules[moduleNameChangeDetectedIn];
       const oneAppDevCdnAddress = `http://${getIp()}:${process.env.HTTP_ONE_APP_DEV_CDN_PORT || 3001}`;
 
       moduleData.browser.url = moduleData.browser.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
       moduleData.legacyBrowser.url = moduleData.legacyBrowser.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
       moduleData.node.url = moduleData.node.url.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
+      moduleData.baseUrl = moduleData.baseUrl.replace('[one-app-dev-cdn-url]', oneAppDevCdnAddress);
```
smackfu, line 53: Is this line tested at all?

10xLaCroixDrinker (approved):

PixnBits:
this is also addresed in #1370

src/server/utils/watchLocalModules.js
Matthew-Mallimo: Should be covered by the existing test cases that Michael modified

Matthew-Mallimo (approved):

src/server/utils/watchLocalModules.js
PixnBits: were they failing before?

